PR #1, opened 2019-02-26
baremetal: add some notes on next steps

markmc:
I want to capture the thoughts from metalkube/dev-scripts#93

### DIFF
--- a/docs/dev/baremetal.md
+++ b/docs/dev/baremetal.md
@@ -35,6 +35,31 @@ Forking is painful, but ...
   openshift/installer, it's quite beneficial for us to be working in
   that codebase from the start.
 
+## What's next?
+
+### Bootstrap Ignition Customizations
+
+We need to make bare metal specific bootstrap ignition customizations.
+How best to do it? With per platform directories for assets?
+
+### Provisioning Master Nodes
+
+How do we prototype the Ironic based provisioning initially? With a
+path that is completely different from the other platforms?
+
+Would we move to terraform next as something more workable in the
+medium term, since that's what other platforms do?
+
+When we would switch to using baremetal-operator? Would all other
+platforms have to switch from terraform at the same time, or could it
+be bare metal specific initially?
+
+### Using External Ironic
+
+Users may use Ironic to discover and introspect nodes before launching
+kni-install. Does it make sense for kni-install to reuse this rather
+than re-running Ironic on the bootstrap VM.
+
 ## How to rebase?
 
 The kni-installer rebasing branch regularly rebases to latest


docs/dev/baremetal.md
```diff
@@ -35,6 +35,24 @@ Forking is painful, but ...
   openshift/installer, it's quite beneficial for us to be working in
   that codebase from the start.
 
+## What's next?
+
+### Bootstrap Ignition Customizations
+
+We need to make are metal specific bootstrap ignition customizations.
```
russellb: s/are/our/ ?

docs/dev/baremetal.md
```diff
@@ -35,6 +35,24 @@ Forking is painful, but ...
...
+How best to do it? With per platform directories for assets?
+
+### Provisioning Master Nodes
+
+Do we use baremetal-operator to drive the provisioning? Do this via
+terrarorm or directly?
```
russellb: I'm -1 on the operator since it seems like too big a deviation from the openshift-install infrastructure.

I'd lean toward directly, or via terraform (another variation on directly).  Neither seem straight forward.  The terraform path fits in much cleaner, but requires us to write a bunch of terraform provider code.  Direct seems easier, but is very different from the other platforms, which are all terraform, so that might be an ugly integration.

russellb (commented):
not sure if you were looking for discussion on the content, but here you go :-)

docs/dev/baremetal.md
markmc: bare!

docs/dev/baremetal.md
markmc: I'm thinking about this a bit flippantly, honestly - whatever hack we were going to do in dev-scripts, let's hack that into kni-installer first. Like, instead of doing terraform.Apply() for the masters, spawn a shell script! No doubt the devil is in the details, I just mean we don't have to make the first cut of this closer perfect just because we're doing it golang.

markmc: Let me know if the latest rev expresses that a bit better

russellb (approved):
lgtm

docs/dev/baremetal.md
hardys: Yeah the gap here is just some way to drive Ironic to deploy the masters - support is going into gophercloud so we should be able to do this in golang, and potentially via Terraform, but for now I'm fine with a hack to get us started, atm we're just driving openstackclient via a script, which works.

Unless openshift-install is moving away from Terraform imminently, I think the right plan is to use terraform, which in theory should be simple-ish as there are far fewer resources to create than in the AWS/OpenStack case, but that assumes we can encapsulate the workflow to deploy nodes in terraform-provider-openstack which may or may not be acceptable to the maintainers (or, we create our own terraform provider which might be an easier first step)

docs/dev/baremetal.md
hardys: :bear: :metal:

docs/dev/baremetal.md
```diff
@@ -35,6 +35,31 @@ Forking is painful, but ...
   openshift/installer, it's quite beneficial for us to be working in
   that codebase from the start.
 
+## What's next?
+
+### Bootstrap Ignition Customizations
+
+We need to make bare metal specific bootstrap ignition customizations.
+How best to do it? With per platform directories for assets?
```
hardys, line 43: @celebdor was going to look into this - I'm not sure static files are enough in all cases as the current dev-scripts approach does some templating.  It'd be nice to avoid doing much stuff in the golang assets though, it's kinda ugly writing go code to generate scripts etc, although there is some of that in openshift-install already..

markmc:
@shardy all really useful thoughts - can you capture them in the doc? thanks

hardys:
Yup I'll give it some more thought and propose a patch with some more details.  Tempted to just write a terraform provider we can use to provision the nodes but that may have to wait until next week ;)

PR #2, opened 2019-02-27
baremetal: add recent baremetal keepalived changes

markmc:
Add changes from metalkube/dev-scripts#89 in dev-scripts.


celebdor:
looks good

PR #3, opened 2019-02-27
WIP: baremetal: sketch out how we'd add an API VIP option

markmc:
This is just a sketch of how we'd add options like the API VIP - looks like we don't need this yet though since we're only launching the bootstrap VM

### DIFF
--- a/pkg/asset/installconfig/baremetal/baremetal.go
+++ b/pkg/asset/installconfig/baremetal/baremetal.go
@@ -26,8 +26,25 @@ func Platform() (*baremetal.Platform, error) {
 		return nil, err
 	}
 
+	// FIXME: bare metal - use net.IP type
+	var masterVIP string
+	err = survey.Ask([]*survey.Question{
+		{
+			Prompt: &survey.Input{
+				Message: "Master API VIP",
+				Help:    "The virtual IP address to use for the kubernetes API.",
+				Default: "", // FIXME: bare metal - add a default
+			},
+			Validate: survey.ComposeValidators(survey.Required, ipValidator),
+		},
+	}, &masterVIP)
+	if err != nil {
+		return nil, err
+	}
+
 	return &baremetal.Platform{
-		URI: uri,
+		URI:       uri,
+		MasterVIP: masterVIP,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/types/baremetal/platform.go
```diff
@@ -9,6 +9,11 @@ type Platform struct {
 	// Default is qemu:///system
 	URI string `json:"URI,omitempty"`
 
+	// MasterVIP is the VIP for the master API.
+	// +optional
+	// FIXME: bare metal - add a default, use net.IP type
+	MasterVIP string `json:"masterVIP,omitempty"`
```
vrutkovs, line 15: This probably should go to a dedicated `network` config

pkg/asset/installconfig/baremetal/baremetal.go
```diff
@@ -26,8 +26,25 @@ func Platform() (*baremetal.Platform, error) {
 		return nil, err
 	}
 
+	// FIXME: bare metal - use net.IP type
+	var masterVIP string
+	err = survey.Ask([]*survey.Question{
```
celebdor, line 31: The philosophy of the openshift installer is to reduce the amount of survey questions as much as possible.

Since one of the requirements we impose on the admins is that they set up a DNS record for api.clustername.basedomain. I think the best to do here is to get it using the net library like in the example below:

```
package main

import (
        "net"
        "fmt"
        "os"
)

func main() {
        domain := "api.ostest.test.metalkube.org"
        ips, err := net.LookupIP(domain)
        if err != nil {
                fmt.Fprintf(os.Stderr, "Could not get IPs: %v\n", err)
                os.Exit(1)
        }
        for _, ip := range ips {
                fmt.Printf("%s IN A %s\n", domain, ip.String())
        }
}```

pkg/types/baremetal/platform.go
celebdor: Do you mean the networkConfig? I think I agree with Mark here on having it as platform install config.

pkg/asset/installconfig/baremetal/baremetal.go
celebdor: After discussing it with @russellb. It's better not to do what I propose so that we can work in environments where the DNS record points to a preNATted address.

stbenjam:
Is this needed anymore in light of https://github.com/openshift-metal3/kni-installer/pull/82?

markmc:
> Is this needed anymore in light of #82?

I guess not

PR #6, opened 2019-03-01
baremetal: add docs on rebasing to latest upstream

markmc:
My attempt to document a repeatable procedure for rebasing.

This is definitely complicated but the goal is simple:

1. Everyone should be able to develop on the master branch without worry
2. We should separately maintain a branch which can eventually become a patch series submitted upstream

So the master branch never rebases, and all our changes go in there via pull requests. But a small number of maintainers can own the job of looking after the rebasing branch, and updating both branches with latest upstream.

### DIFF
--- a/docs/dev/baremetal.md
+++ b/docs/dev/baremetal.md
@@ -62,23 +62,76 @@ than re-running Ironic on the bootstrap VM.
 
 ## How to rebase?
 
-The kni-installer rebasing branch regularly rebases to latest
+The kni-installer `rebasing` branch regularly rebases to latest
 openshift/installer whereas the master branch will never rebase and
 only merges from openshift/installer.
 
-So, the procedure is roughly:
+### Prepare the repo
 
-FIXME: add instructions for rebasing any changes on kni-installer
-master onto the rebasing branch, and then tidying up the patch
-series.
+Checkout this repo and add `github.com/openshift/installer` as a
+remote that we'll call `upstream`:
 
-- Checkout the rebasing branch and find the "Rename to kni-install"
-  commit. Let's call that $RENAME_COMMIT.
-- Fetch latest openshift/installer - let's say you've named its
-  git remote "upstream". It's a good idea to familiarize yourself
-  with the changes since your last rebase.
-- Create a temporary branch from latest openshift/install and
-  redo the rename:
+```sh
+git clone git@github.com:metalkube/kni-installer.git
+cd kni-installer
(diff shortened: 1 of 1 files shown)


hardys:
This does look a little complex, interested to understand the rationale here vs having, say, a baremetal_patches branch which we can periodically rebase on master?

I don't have a strong opinion either way, but personally I'm more familiar with maintaining some feature/test branch that's rebased on a moving master, and that might translate more naturally to openshift-install PRs when we're ready to do that?

Maybe the branch that gets rebased doesn't matter, I'm just wondering if it'd be clearer to have a specific set of patches which we rebase vs periodic merge commits from the rebasing branch?

markmc:
> This does look a little complex,

Believe it or not, this is roughly the approach Stephen Tweedie came up with when we were maintaining out-of-tree kernel support for Xen dom0 ... and ultimately we wanted to get that patch series merged upstream

> interested to understand the rationale here vs having, say, a baremetal_patches branch which we can periodically rebase on master?

Ok, so the master branch would just mirror openshift/installer master and kni-installer would live on a rebasing branch

The problem with this is the impact that a rebase has on anyone who is contributing to kni-installer

So, for example, if you had 5 pull requests open against kni-installer and then you rebased the branch, you'd have to rebase all the pull requests onto the new branch

A fairly widely accepted principle is that if you have multiple people working on a branch, you should never rebase it - so we'd be very clear who is looking after the rebasing branch, and everyone else can just work away on the non-rebasing branch

> I don't have a strong opinion either way, but personally I'm more familiar with maintaining some feature/test branch that's rebased on a moving master,

With a lot of people working together on that rebasing feature branch?

I mean, maybe the impact on contributors of rebasing wouldn't be that large ... but I know there's one camp that would say you should never rebase a branch like this

> and that might translate more naturally to openshift-install PRs when we're ready to do that?
>
> Maybe the branch that gets rebased doesn't matter, I'm just wondering if it'd be clearer to have a specific set of patches which we rebase vs periodic merge commits from the rebasing branch?

Yes, that's the goal of keeping a rebasing branch

On master, you'll see merge commits from openshift/installer master and merges of changes from PRs to kni-installer

On the rebasing branch, you'll just see a clean patch series, eventually suitable for sending as PRs to openshift/installer

hardys:
Ack thanks for the additional info, makes sense so lets merge and iterate on any further details when we get more comfortable with the process.  Maybe we can work out a regular cadence for the rebases e.g at the start of each development sprint or something.

PR #7, opened 2019-03-01
baremetal: prefix resources with cluster_id

markmc:
As per openshift/installer#1280 cluster_id is now used to prefix resources in other providers rather than cluster_name.


markmc:
Today's rebase is completely broken without this, so I'm going to go ahead and merge

PR #8, opened 2019-03-01
baremetal: avoid separate base image for bootstrap

markmc:
The libvirt platform driver creates a base volume, and then creates
copy-on-write volumes for various VMs it creates.

The baremetal platform driver only creates a single libvirt VM, so it
does not need this additional layer.

This fixes "destroy bootstrap" not cleaning up the base image. The
base image resource is owned by module.volume, but "destroy boostrap"
only destroys the resources in module.bootstrap. This makes sense for
the libvirt platform, but not for the baremetal platform.


hardys:
Tested and this works well, perhaps we might add something to the dev-scripts cleanup script as I had to clean the stale images manually (there were a few due to the new random component of the bootstrap image name and repeated testing if dev-scripts #93)

PR #9, opened 2019-03-05
baremetal: fix keepalived template name

markmc:
We couldn't use .template because the bootstrap ignition asset
rendering code will automatically render .template files with data
like the pull secret (search for bootstrapTemplateData).

Just use a .tmpl suffix to avoid this.


markmc:
Merging this - reverting an obvious screwup in #2

PR #10, opened 2019-03-05
baremetal: enable keepalived

markmc:
Turns out there's a hardcoded list of systemd units to enable.


markmc:
We're broken without this straightforward change, so merging

PR #12, opened 2019-03-11
docs: add kni-install info to top-level README

markmc:
This will be a bit annoying if it causes rebasing pain, but it will help people who navigate to the repo without understanding the context.


markmc:
Going to merge this quickly to resolve feedback I received about confusion. Happy to follow-up on any suggestions for improvements though.

PR #13, opened 2019-03-12
Make sure we only return a single dev

derekhiggins:
In cases where a nic has two ipaddresses on the same externel
subnet then "ip -o addr show" returns a row for each one. Make
sure we only attempt to use one.

Was hitting a problem on a bootstrap not that has 2 IPs on the same subnet


data/data/bootstrap/files/usr/local/bin/keepalived.sh
```diff
@@ -35,7 +35,7 @@ for iface_cidr in iface_cidrs:
 sys.exit(1)
 EOF
 )
-    ip -o addr show to "$net_cidr" | awk '{print $2}'
+    ip -o addr show to "$net_cidr" | head -n 1 | awk '{print $2}'
```
celebdor: I would really prefer if we have some stricter logic than just choosing whichever address was configured first for the link. It should be fine to merge this in the meantime.

celebdor (approved):

markmc:
Oh, @celebdor approved. Merging ...

PR #14, opened 2019-03-13
Latest openshift install

markmc:
Pointed out by @vrutkovs that a rebase might be needed. I haven't tested this yet.


markmc:
I think we're good to go with this

PR #15, opened 2019-03-13
baremetal: move to openshift-metalkube org

markmc:
The kni-installer repo is moving from the metalkube org to the  openshift-metalkube org.


russellb:
lgtm

PR #20, opened 2019-03-16
bootstrap resolv.conf to use own coredns-mdns

celebdor:
In order for the clustering to work, the bootstrap node should be able
to resolve the SRV records for Etcd. In order to do that, the best is to
prepend localhost to /etc/resolv.conf so that as soon coredns-mdns is
ready, we will be able to use it to get the master entries.


celebdor:
Needs to be tested with https://github.com/openshift-metalkube/dev-scripts/pull/171

vrutkovs:
LGTM

hardys (approved):
Tested with dev-scripts #171 and now working well for me :)

PR #24, opened 2019-03-21
Pull in terraform-provider-ironic

stbenjam:
This PR is vendoring only, actual master deployment PR coming shortly. I wanted to separate this from my code to make it easier to review.

This pulls in terraform-provider-ironic, which required a newer gophercloud and newer terraform-provider-openstack.


stbenjam:
Also: If/when openshift/installer ever updates its deps it's going to create a mess when we rebase.

stbenjam:
Probably not needed, github keeps it the vendored stuff relatively separate.

PR #25, opened 2019-03-22
Deploy masters with kni-installer

stbenjam:
For use with https://github.com/openshift-metalkube/dev-scripts/pull/200 -- replaces `07_deploy_masters.sh` and deploys the cluster with the installer.


data/data/baremetal/masters/main.tf
```diff
@@ -0,0 +1,115 @@
...
+    image_source   = "${var.image_source}"
+    image_checksum = "${var.image_checksum}"
+    root_gb        = "${var.root_gb}"
+    root_device    = "${var.root_disk}"
+  }
+
+  driver = "ipmi"
+
+  driver_info {
+    ipmi_port      = "${var.nodes["nodes_0_driver_info_ipmi_port"]}"
```
stbenjam: I think this is a concern for the bare metal testers who didn't want to specify a port.  This whole section is also hard-coded to IPMI.  It might be a good idea for me to see if I can't extract driver_info into it's own map so users can specify whatever they want - ipmi, idrac, etc.

dhellmann: The code in https://github.com/metalkube/baremetal-operator/blob/master/pkg/bmc/access.go (as well as idrac.go and ipmi.go in that directory) may be useful, at least as a reference. Notice that the driver_info *keys* change based on the driver.

stbenjam: Yes, that's helpful - thanks. 

The main problem I have is terraform, it doesn't have very robust data structures.  A list can only be strings.  A map can only be string/string pairs.  I have to think about how a user is going to represent the hardware info in installer-config.yaml for kni-installer.  The easiest way is something like this

```
driver_info_master_0:
  drac_username: admin
  drac_password: potato
  drac_address: ...
driver_info_master_1:
  drac_username: admin
  drac_password: potato
  drac_address: ...
```

But then you have to know in advance how masters there are for the terraform template.  It's very limiting.

dhellmann: I would stay away from "drac" as a prefix since as soon as we have some other hardware support in place it will look odd. BMC would work. The address field can be a URL like idrac+https://host:port/path so we get the type from that.

The username and passwords will end up going into a secret (one per host) so using separate fields will save you having to parse them out of the URL yourself.

So, the format you have looks OK to me.

vrutkovs (approved):

markmc:
This all looks pretty good and I'm keen to see it merged ASAP, leaving any improvements to further iterations. I'm going to play with this for a bit myself now.

The main thing that needs more thought and work as a future iteration is the way we passed the nodes data in install-config.yaml - @stbenjam calls this out in the dev doc too

hardys:
Ok, dev-scripts/#200 passed CI, and it's been painful to rebase so given the comments above lets merge this and iterate on the install-config in a future PR

PR #28, opened 2019-03-28
Seed the random number generator

derekhiggins:
Ensure operations requiring random data get it, calls
to randomMACAddress() were providing the same mac address
in multiple environments and in some cases getting the same
IP address if not isolated.

### DIFF
--- a/cmd/kni-install/main.go
+++ b/cmd/kni-install/main.go
@@ -3,9 +3,11 @@ package main
 import (
 	"flag"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


derekhiggins:
Submitted upstream but also submitting here incase that takes time to resolve
https://github.com/openshift/installer/pull/1478

hardys:
Should we merge this since the upstream patch seems to be taking a while to discuss/review?

markmc:
Looks like @derekhiggins has feedback that this should be initialized in terraform-provider-libvirt instead - so we can carry this for now, but I expect something different will ultimately be merged upstream

PR #31, opened 2019-04-02
Avoid using the subnet.2 ip

e-minguez:
The nthhost file doesn't seem to be used anywhere else, so I think it is safe to remove it.


celebdor:
I distilled the commit into the actual fix at https://github.com/openshift-metalkube/kni-installer/pull/33 keeping the commit owner.

PR #39, opened 2019-04-04
bootkube.sh Remove duplicate MCO rendering

celebdor:
As part of the rebase, we ended up with the MCO rendering being done twice.


stbenjam (approved):
Lgtm

markmc:
Looks correct to me - in our fork, we need MCO rendering to happen earlier and a merge conflict caused us to add a duplicate in its original location and the moved code didn't get the etcd-metric-ca-bundle.crt change

If there's any way we can submit the MCO rendering reording change upstream ASAP to openshift/installer that would help a lot

stbenjam:
Sorry about that, when resolving the conflict I didn't notice the duplicate MCO blocks. :frowning_face:

PR #43, opened 2019-04-09
Update bootkube output when waiting on DNS SRV.

russellb:
This log output doesn't actually show up in the log while we're in the
middle of waiting because of the lack of a newline.  Change this so if
you're following bootkube, you see this message as soon as we start
this section.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
@@ -134,15 +134,14 @@ then
 	touch mco-bootstrap.done
 fi
 
-echo -n "Waiting for etcd member names to be available from SRV"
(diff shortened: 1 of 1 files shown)


russellb:
From my latest deployment, for example, we get this line in the log once this section completes:

```
Apr 09 14:06:48 localhost.localdomain bootkube.sh[1446]: Waiting for etcd member names to be available from SRV.................................................................................................................................................................................................................................................................................................................................................................................................. Found etcd members!
```

celebdor (approved):
It's a bit sad to have to take so many vertical lines. Anyway, logs are important :-)

celebdor (approved):

russellb:
It won't take up a bunch of lines, since I removed the `.` printed every second.

PR #47, opened 2019-04-16
Rewrite terraform to support dynamic number of nodes

yrobla:
Change the data structure to accept data of nodes using maps.
This will allow to add a variable number of masters (1,3) in the future.
Added a new dependency on rodaine/hclencoder, to render terraform data
in HCL instead of JSON, to avoid a bug in JSON parsing:
https://github.com/hashicorp/terraform/issues/15549


data/data/baremetal/variables-baremetal.tf
```diff
@@ -1,91 +1,49 @@
-variable "ironic_uri" {
+variable "IronicURI" {
```
stbenjam: nit: these need to stay camel case (`ironic_uri`) according to terraform style guide - https://github.com/jonbrouse/terraform-style-guide#parameter-meta-parameter-and-variable-naming.  

We want to get all this in openshift/installer and they're a bit picky

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -17,21 +17,11 @@ type config struct {
...
-	RootDevice1 interface{} `json:"root_device_1"`
-	DriverInfo1 interface{} `json:"driver_info_1"`
-
-	Master2     interface{} `json:"master_2"`
-	Properties2 interface{} `json:"properties_2"`
-	RootDevice2 interface{} `json:"root_device_2"`
-	DriverInfo2 interface{} `json:"driver_info_2"`
+    // limitation that maps cannot be strings
+	MasterNodes     interface{} `json:"master_nodes"`
+    Properties      interface{} `json:"properties"`
```
stbenjam: nit: need to run `go fmt` on all the go code

data/data/baremetal/main.tf
```diff
@@ -1,44 +1,31 @@
-provider "libvirt" {
-  uri = "${var.libvirt_uri}"
-}
+resource "ironic_node_v1" "openshift-master" {
```
stbenjam: I think this belongs in baremetal/masters, the base main.tf sets up the providers and modules

stbenjam (commented):
I get some errors trying to run this - it looks like the `main.tf` in the root of the directory was overwritten by the one that should be in `masters/`

```
level=error msg="Error: resource 'ironic_node_v1.openshift-master' count: unknown variable referenced: 'master_nodes'; define it with a 'variable' block"
level=error
level=error
level=error
level=error msg="Error: resource 'ironic_node_v1.openshift-master' config: unknown variable referenced: 'image_source'; define it with a 'variable' block"
level=error
level=error
level=error
level=error msg="Error: resource 'ironic_node_v1.openshift-master' config: unknown variable referenced: 'root_gb'; define it with a 'variable' block"
level=error
```

data/data/baremetal/variables-baremetal.tf
yrobla: So that's part of the hcl renderer, not sure on how to change it. I had to stop using JSON renderer because of the referenced bug. Instead, i used that project that renders to HCL format automatically, but it generates the vars with upper case... what should i do there?

data/data/baremetal/main.tf
yrobla: There was a wrong file copy there

data/data/baremetal/variables-baremetal.tf
stbenjam: I think you should be able to use the annotation support in go, `hcl:"ironic_uri"`

stbenjam: See the example here: https://github.com/rodaine/hclencoder

data/data/baremetal/main.tf
```diff
@@ -21,24 +21,14 @@ module "masters" {
...
-  driver_info_1 = "${var.driver_info_1}"
-
-  master_2 = "${var.master_2}"
-  properties_2 = "${var.properties_2}"
-  root_device_2 = "${var.root_device_2}"
-  driver_info_2 = "${var.driver_info_2}"
+  image_source   = "${var.master_configuration["ImageSource"]}"
+  image_checksum = "${var.master_configuration["ImageChecksum"]}"
+  root_gb        = "${var.master_configuration["RootGb"]}"
+  root_disk      = "${var.master_configuration["RootDisk"]}"
```
stbenjam: Install is failing since these don't get set (dev-scripts is using image_source, etc)

stbenjam: Actually my install is still failing for some other reason, but you should be able to set the HCL names for these in `pkg/types/baremetal/platform.go`:

```go
type MasterConfiguration struct {
        ImageSource   string `hcl:"image_source" json:"image_source,omitempty"`
        ImageChecksum string `hcl:"image_checksum" json:"image_checksum,omitempty"`
        RootGb        string `hcl:"root_gb" json:"root_gb,omitempty"`
}
```

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -2,42 +2,30 @@
...
-	RootDevice1 interface{} `json:"root_device_1"`
-	DriverInfo1 interface{} `json:"driver_info_1"`
-
-	Master2     interface{} `json:"master_2"`
-	Properties2 interface{} `json:"properties_2"`
-	RootDevice2 interface{} `json:"root_device_2"`
-	DriverInfo2 interface{} `json:"driver_info_2"`
+    // limitation that maps cannot be strings
+	MasterNodes     interface{} `hcl:"master_nodes"`
+    Properties      interface{} `hcl:"properties"`
```
stbenjam: Can you run `go fmt` again?

stbenjam (commented):
Otherwise this looks good.  I get past the masters install so I think this is all working

stbenjam (approved):

hardys (approved):
Tested locally and this works for me with the default 3 masters, not tested with 1 yet

PR #52, opened 2019-04-20
Extend timeout for baremetal deployment

karmab:


stbenjam:
@karmab I'd like to close this; we won't be able to bring this change to `openshift/installer`, is that ok?

If it's really needed we'll have to work to convince them to take this there.  We would need to gather data to show that baremetal is special and needs a longer time out.  For bootstrap, though, we had success last week even when bootstrap is running in a layer of nested virtualization, as long as we're using VirtIO.

stbenjam:
Please re-open if it is an issue we need to address; but to convince openshift/installer, we're going to need to have some empirical data and a reproducer that can show why we need longer than their default timeouts for anything.

PR #55, opened 2019-04-23
Work around extra host route.

russellb:
This is the kni-installer version of the following commit:
https://github.com/openshift-metalkube/dev-scripts/commit/21cb51d7e2476824c8571fc73ebbdb003bd8df19

The dev-scripts commit fixed the issue for masters, but the issue can
still occur on the bootstrap VM, breaking the deployment process.

The issue is that there's a /32 route someimtes not getting deleted
when the VIP moves to another node.  This changes the keepalived
config such that the route created matches another route that is still
correct for the host.


russellb:
related dev-scripts PR: https://github.com/openshift-metalkube/dev-scripts/pull/377

original dev-scripts issue: https://github.com/openshift-metalkube/dev-scripts/issues/344

follow-up kernel bug: https://bugzilla.redhat.com/show_bug.cgi?id=1700415

celebdor (approved):

PR #51, opened 2019-04-20
Make bootstrap work on flexible dns entries

yrobla:
Depending on having 1 or 3 masters, the number of dns entries
we need to wait for is different. We can check the number of masters
in cluster-config.yaml, and parameterize the pod on bootstrap,
to wait for the needed dns members.

PR #71, opened 2019-05-02
bootstrap to be dns vip master until the end

celebdor:
The reason for this is that during bootstrapping, and only during
bootstrapping we want to enforce that unless TOTAL_AMOUNT_OF_MASTERS is
available for the Etcd SRV record, we give out NXDOMAIN answer for the
SRV query that Etcd uses to cluster.

Signed-off-by: Antoni Segura Puimedon <antoni@redhat.com>


celebdor:
Should be rebased after https://github.com/openshift-metal3/kni-installer/pull/51

celebdor:
Goes together with https://github.com/openshift-metal3/dev-scripts/pull/501

PR #77, opened 2019-05-10
Add api-int record to bootstrap

cybertron:
Add the hosts plugin to coredns so that we can create a static entry
for api-int.$CLUSTER_DOMAIN. hosts is used because it doesn't have to
be authoritative for the zone and can allow fallthrough of records
that are not found.

### DIFF
--- a/data/data/bootstrap/files/etc/coredns/Corefile
+++ b/data/data/bootstrap/files/etc/coredns/Corefile
@@ -5,4 +5,7 @@
     forward . /etc/coredns/resolv.conf
     cache 30
     reload
+    hosts /etc/coredns/api-int.hosts {$CLUSTER_DOMAIN} {
+        fallthrough
+    }
 }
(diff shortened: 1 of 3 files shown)


cybertron:
@celebdor This should get api-int working on the bootstrap node too.

stbenjam (approved):
Master nodes do get their ignition successfully with this change.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/632/

stbenjam (changes requested):
D'oh, sorry - I had my change from this morning still applied - https://github.com/openshift-metal3/dev-scripts/pull/531 :-(

With just this, it's not enough.  Ignition hosts aren't pointing at coredns, they're all hitting 192.168.111.1, which doesn't have the api-int record.

stbenjam:
I added the temporary record to https://github.com/openshift-metal3/dev-scripts/pull/523.  I suppose we still need this as well, for when after the cluster comes up.

stbenjam (approved):

PR #82, opened 2019-05-17
Point Ignition at API VIP directly

cybertron:
Instead of having Ignition use the DNS name for the API, which will
be problematic in 4.1 where it switches to api-int, get the VIP from
the configuration and use that directly.

### DIFF
--- a/pkg/asset/ignition/machine/node.go
+++ b/pkg/asset/ignition/machine/node.go
@@ -13,6 +13,12 @@ import (
 // pointerIgnitionConfig generates a config which references the remote config
 // served by the machine config server.
 func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, role string) *ignition.Config {
+	ignitionHost := fmt.Sprintf("api-int.%s:22623", installConfig.ClusterDomain())
+	if installConfig.Platform.BareMetal != nil {
+		// Baremetal needs to point directly at the VIP because we don't have a
+		// way to configure DNS before Ignition runs.
+		ignitionHost = fmt.Sprintf("%s:22623", installConfig.BareMetal.ApiVIP)
+	}
 	return &ignition.Config{
 		Ignition: ignition.Ignition{
 			Version: ignition.MaxVersion.String(),
@@ -21,7 +27,7 @@ func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, ro
 					Source: func() *url.URL {
 						return &url.URL{
 							Scheme: "https",
-							Host:   fmt.Sprintf("api-int.%s:22623", installConfig.ClusterDomain()),
+							Host:   ignitionHost,
 							Path:   fmt.Sprintf("/config/%s", role),
 						}
 					}().String(),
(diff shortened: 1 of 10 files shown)


pkg/asset/ignition/machine/node.go
```diff
@@ -21,7 +21,7 @@ func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, ro
 					Source: func() *url.URL {
 						return &url.URL{
 							Scheme: "https",
-							Host:   fmt.Sprintf("api-int.%s:22623", installConfig.ClusterDomain()),
+							Host:   fmt.Sprintf("%s:22623", installConfig.BareMetal.ApiVIP),
```
stbenjam: We need a solution that we can bring to openshift/installer, can we add a generic per-platform configuration which defaults to api-int except for baremetal?

russellb: Isn't this changing generic code, and not baremetal platform specific?

russellb (dismissed):

pkg/asset/ignition/machine/node.go
cybertron: Yes. The current version of this is based on Roy's work in https://github.com/rgolangh/installer/commit/69f973204473e567aa599066228b7751df745c91 which changed common code. I need to look at how to make it upstreamable.

stbenjam: Also, just asking, but couldn't we use `api.${CLUSTER_DOMAIN}` instead of the VIP for baremetal? Then no need to add IP's to certs, etc.

cybertron: No, api-int was added because we couldn't use api for internal communication. It had to do with possible user-supplied certificates on the api endpoint that wouldn't be trusted by our nodes. There's more discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1685704

pkg/asset/ignition/machine/node.go
```diff
@@ -13,6 +13,12 @@ import (
 // pointerIgnitionConfig generates a config which references the remote config
 // served by the machine config server.
 func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, role string) *ignition.Config {
+	ignitionHost := fmt.Sprintf("api-int.%s:22623", installConfig.ClusterDomain())
+	if installConfig.Platform.BareMetal != nil {
```
cybertron, line 17: I found a couple of other instances of platform checks in common code so I'm hoping this is okay. I'm open to other suggestions though.

pkg/asset/ignition/machine/node.go
celebdor: Can't we just put some ifs?

cybertron: Yeah, that's what I ended up doing.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/648/

stbenjam (approved):
CI results aren't actually valid until https://github.com/openshift-metal3/dev-scripts/pull/558 makes it in. But, this PR looks good to me.

cybertron:
I think this actually requires a change to dev-scripts before it will work. We need to add api_vip to install-config.yaml to reflect the new survey question added here. I'll get a PR proposed to do that.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/672/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/676/

PR #88, opened 2019-05-23
Detect VIP from DNS when not provided in survey

cybertron:


stbenjam:
Does this supersede https://github.com/openshift-metal3/kni-installer/pull/82? Can you close that one then?

cybertron:
I'm not sure the VIP lookup can merge in its current form. I wanted to get this commit up for discussion, but if #82 is ready to go we can merge that alone and then I'll just rebase this.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/680/

cybertron:
Rebased this and added a validation so we abort if the DNS lookup fails.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/710/

pkg/types/baremetal/defaults/platform.go
```diff
@@ -35,4 +40,15 @@ func SetPlatformDefaults(p *baremetal.Platform) {
 			host.HardwareProfile = HardwareProfile
 		}
 	}
+
+	if p.ApiVIP == ApiVIP {
```
stbenjam, line 44: nit: `if p.ApiVIP == ""` would be clearer to me what's going on

cybertron: I changed it to use the const because it is also referenced in the baremetal installconfig survey validation and I wanted to make sure the values matched, even if we changed the default for some reason. I do agree that referencing it from within the package is confusing though, so I'd be okay changing it back.

stbenjam: Oh I see, that makes sense to me, it's fine then

stbenjam (approved):
LGTM

PR #90, opened 2019-05-29
Provide BareMetalMachineProviderSpec with the right fields

### LINKED ISSUE #87: Provide the BareMetalMachineProviderSpec with the right fields

sadasu:
Fixes #87 

kni-install will create the master Machines and worker MachineSet
with the baremetal machine providerSpec. cluster-api-provider-baremetal
expects two fields: the image and userdata (checksum URL).
After this change the cluster-api-provider baremetal can now start
using this image provided via the BareMetalMachineProviderSpec instead
of a hard-coded one.

### DIFF
--- a/pkg/asset/machines/baremetal/machines.go
+++ b/pkg/asset/machines/baremetal/machines.go
@@ -9,6 +9,7 @@ import (
 	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+        corev1 "k8s.io/api/core/v1"
 
 	"github.com/openshift-metalkube/kni-installer/pkg/types"
 	"github.com/openshift-metalkube/kni-installer/pkg/types/baremetal"
@@ -60,5 +61,11 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 }
 
 func provider(clusterName string, networkInterfaceAddress string, platform *baremetal.Platform, userDataSecret string) *baremetalprovider.BareMetalMachineProviderSpec {
-	return &baremetalprovider.BareMetalMachineProviderSpec{}
+	return &baremetalprovider.BareMetalMachineProviderSpec{
+                Image: baremetalprovider.Image{
+                        URL: platform.MasterConfiguration["image_source"].(string),
+                        Checksum: platform.MasterConfiguration["image_checksum"].(string),
+                },
+                UserData: &corev1.SecretReference{Name: userDataSecret},
+        }
 }


dhellmann:
During the initial installation, we're going to need an image source URL that points to the bootstrap node. Once the cluster is up, the Machine and MachineSet values should have a URL that points to the image hosting service running inside the cluster. I don't know if that means we need 2 separate configuration values, or if the value being passed via this change is going to be the correct "day 2" value. What sets the image_source configuration option, and what value is used?

sadasu:
Referring to this : https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/docs/api.md it appears that the values being passed into BareMetalMachineProviderSpec are for day 2. 

There is another image in the install-config which can be seen here: https://github.com/openshift-metal3/kni-installer/blob/master/data/data/baremetal/main.tf#L14 and https://github.com/openshift-metal3/kni-installer/blob/master/pkg/tfvars/baremetal/baremetal.go#L37. I am not sure if this is the image to be used in the bootstrap node. 

Looking at ocp/install-config.yaml.tmp which is the install-config that is currently being generated by dev-scripts, the value of image source is 
image_source: "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2"

dhellmann:
> Referring to this : https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/docs/api.md it appears that the values being passed into BareMetalMachineProviderSpec are for day 2.
> 
> There is another image in the install-config which can be seen here: https://github.com/openshift-metal3/kni-installer/blob/master/data/data/baremetal/main.tf#L14 and https://github.com/openshift-metal3/kni-installer/blob/master/pkg/tfvars/baremetal/baremetal.go#L37. I am not sure if this is the image to be used in the bootstrap node.

I think that's the image for the bootstrap node, but not the image that the bootstrap node should use to deploy the control plane nodes.

> 
> Looking at ocp/install-config.yaml.tmp which is the install-config that is currently being generated by dev-scripts, the value of image source is
> image_source: "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2"

OK. That value is only valid for day 1.

I think your change here is good, but that we will have to change the value passed in by the installer to reflect the valid URL for day 2. I don't know how we know what that value is, though.

pkg/asset/machines/baremetal/machines.go
```diff
@@ -60,5 +60,10 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 }
 
 func provider(clusterName string, networkInterfaceAddress string, platform *baremetal.Platform, userDataSecret string) *baremetalprovider.BareMetalMachineProviderSpec {
-	return &baremetalprovider.BareMetalMachineProviderSpec{}
+	return &baremetalprovider.BareMetalMachineProviderSpec{
+                Image: baremetalprovider.Image{
+                        URL: platform.MasterConfiguration["image_source"].(string),
+                        Checksum: platform.MasterConfiguration["image_checksum"].(string),
+                },
```
russellb, line 68: need to set the user data field, as well

russellb (changes requested):

imain:
> > Referring to this : https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/docs/api.md it appears that the values being passed into BareMetalMachineProviderSpec are for day 2.
> > There is another image in the install-config which can be seen here: https://github.com/openshift-metal3/kni-installer/blob/master/data/data/baremetal/main.tf#L14 and https://github.com/openshift-metal3/kni-installer/blob/master/pkg/tfvars/baremetal/baremetal.go#L37. I am not sure if this is the image to be used in the bootstrap node.
> 
> I think that's the image for the bootstrap node, but not the image that the bootstrap node should use to deploy the control plane nodes.
> 
> > Looking at ocp/install-config.yaml.tmp which is the install-config that is currently being generated by dev-scripts, the value of image source is
> > image_source: "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2"
> 
> OK. That value is only valid for day 1.
> 
> I think your change here is good, but that we will have to change the value passed in by the installer to reflect the valid URL for day 2. I don't know how we know what that value is, though.

Yes I think this is all correct.  I'm not sure where the day 2 URL will come from yet either.  I'm hoping we can set up a k8s service and be able to get the address from kubernetes.  Haven't gotten to exploring this yet. :/

sadasu:
Removing WIP status after testing a few times with success.

russellb:
Thanks!

russellb (approved):

Issue #66, opened 2019-04-30
Undo rename to kni-installer


russellb:
We renamed the fork to kni-installer and updated all of the go import paths to reflect this.  This needs to be undone before we can submit a PR back to openshift/installer.

We can retain this fork, but I imagine a typical env will clone openshift/installer, and then add our fork as a branch, leaving the repo under github.com/openshift/installer in their gopath.

stbenjam:
I undid the rename on a personal [branch](https://github.com/openshift/installer/compare/master...stbenjam:bare-metal#files_bucket) using this:

```console
$ sed -i 's|openshift-metalkube/kni-installer|openshift/installer|g' $(git grep -l openshift-metalkube/kni-installer | grep '\(cmd\|build.sh\|pkg\|assets_generate.go\)')
$ sed -i 's|kni-install|openshift-install|g' $(git grep -l kni-install | grep '\(cmd\|build.sh\|pkg\)')
$ git mv cmd/kni-install cmd/openshift-install
```

What do you think of keeping a branch here with an un-renamed copy of master? For easy comparision with openshift/installer.

russellb:
Is a branch really necessary, though?  Can we just make these changes directly in master?

stbenjam:
Oh, I see - should do, if we move it to `$GOPATH/src/github.com/openshift/installer` everywhere it gets used, otherwise go will be unhappy

markmc:
I made the first patch on the rebasing branch take care of the renaming, so we can easily ignore the noise of the renaming

But maybe we could rename openshift-metalkube/kni-installer to openshift-metalkube/installer and make it so that the only way you build it is like:

```
go get -d github.com/openshift/installer
cd ~/go/src/github.com/openshift/installer
git remote add -f kni git@github.com:openshift-metalkube/installer.git
git checkout -b kni kni/master
```

i.e. you have to build it as a standard branch of openshfit-install

stbenjam:
I think we can just close this, I have a branch here https://github.com/openshift-metal3/installer/pull/1 with the undone rename, which is trivial to redo as needed.

PR #92, opened 2019-05-29
WIP: fix up some dependencies

dhellmann:


Gopkg.toml
```diff
@@ -1,6 +1,6 @@
 ignored = [
-  "github.com/openshift-metalkube/kni-installer/tests*",
-  "github.com/openshift-metalkube/kni-installer/pkg/terraform/exec*"
+  "github.com/openshift-metal3/kni-installer/tests*",
+  "github.com/openshift-metal3/kni-installer/pkg/terraform/exec*"
```
stbenjam, line 3: Did this help? openshift-metalkube/kni-installer should redirect to the repo, and until now I think everything in here is pointing at the old name

dhellmann: No. Now "dep ensure" panics after 30+ minutes. https://gist.github.com/dhellmann/3931a0e1d871ffdbf9df601f1a1fafcf

"dep check" reports a ton of errors, most of them related to parts of the installer importing other parts using the old name: https://gist.github.com/dhellmann/bf663f9a51fc2716d8d649cbb89cbf42

stbenjam: Oh, do you have the installer checked out into `$GOPATH/src/openshift-metal3`? It should be checked out to `$GOPATH/src/openshift-metalkube` until the rename goes through, or we settle on reverting all the name changes alltgoether (#66)

dhellmann:
I don't think we need this after all.

PR #96, opened 2019-06-03
 Merge openshift/installer for 4.2.0-0.ci-2019-06-03-105723 

stbenjam:
Our previous build was GC'd

PR #97, opened 2019-06-03
vendor: update terraform-provider-ironic to v0.1.3

stbenjam:
This should fix most of the warnings from the upgrade to the new version
of terraform.

This should be merged after https://github.com/openshift-metal3/kni-installer/pull/96


dhellmann (approved):
/lgtm

PR #98, opened 2019-06-05
Wait for ironic API to be available

stbenjam:
This change makes terraform wait for the Ironic API to be available
before trying to create any of the Ironic resources. This would permit
us to move Ironic to the bootstrap VM, as otherwise terraform would try
to create the masters and bootstrap VM concurrently.

The mechanism used for this is a null resource in terraform combined
with a local-exec provisioner that polls the API for up to 15 minutes.
The Ironic host records have a depends_on set to this null resource.

### DIFF
--- a/cmd/kni-install/main.go
+++ b/cmd/kni-install/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"github.com/openshift-metalkube/kni-installer/pkg/terraform/exec/provisioners"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -34,12 +35,29 @@ func main() {
 	// https://github.com/openshift/installer/pull/1478
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	if len(os.Args) > 0 {
+	if argLen := len(os.Args); argLen > 0 {
+		// Support external terraform plugins
 		base := filepath.Base(os.Args[0])
 		cname := strings.TrimSuffix(base, filepath.Ext(base))
 		if pluginRunner, ok := plugins.KnownPlugins[cname]; ok {
 			pluginRunner()
 			return
+		}
+
+		// Support internal terraform plugins
+		if argLen > 3 {
+			if os.Args[1] == "internal-plugin" {
+				pluginType := os.Args[2]
+				plugin := os.Args[3]
+
+				switch pluginType {
+				case "provisioner":
+					if provisioner, ok := provisioners.KnownProvisioners[plugin]; ok {
+						provisioner()
+						return
+					}
+				}
+			}
 		}
 	}
 
--- a/data/data/baremetal/masters/main.tf
+++ b/data/data/baremetal/masters/main.tf
@@ -1,3 +1,11 @@
+# This null resource is used to wait for the Ironic API to become available before trying to create
+# the master hosts.
+resource "null_resource" "api-wait" {
+  provisioner "local-exec" {
+    command = "timeout 900 bash -c 'while [[ \"$(curl -s -o /dev/null -w \"%%{http_code}\" ${var.ironic_uri})\" != \"200\" ]]; do sleep 5; done'"
+  }
+}
+
 resource "ironic_node_v1" "openshift-master-host" {
   count          = var.master_count
   name           = var.hosts[count.index]["name"]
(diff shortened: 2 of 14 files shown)


stbenjam:
@hardys FYI

stbenjam:
Seems to be working:

```
Jun 05 16:42:33 |  level=debug msg="module.masters.null_resource.api-wait (local-exec): Executing: [\"/bin/sh\" \"-c\" \"timeout 900 bash -c 'while [[ \\\"$(curl -s -o /dev/null -w \\\"%{http_code}\\\" http://localhost:6385/v1)\\\" != \\\"200\\\" ]]; do sleep 5; done'\"]"
Jun 05 16:42:33 |  level=debug msg="module.masters.null_resource.api-wait: Creation complete after 0s [id=7265670779706103693]"
Jun 05 16:42:33 |  level=debug msg="module.masters.ironic_node_v1.openshift-master-host[2]: Creating..."
Jun 05 16:42:33 |  level=debug msg="module.masters.ironic_node_v1.openshift-master-host[1]: Creating..."
Jun 05 16:42:33 |  level=debug msg="module.masters.ironic_node_v1.openshift-master-host[0]: Creating..."
```

If you want a better test, you can `podman stop ironic` when the installer gets launched, and while `null_resource.api-wait`  is running, do a `podman start ironic`, and you'll see it complete a few seconds later.

cmd/kni-install/main.go
```diff
@@ -34,13 +35,30 @@ func main() {
...
 		}
+
+		// Support internal terraform plugins
+		if argLen > 3 {
+			if os.Args[1] == "internal-plugin" {
+				pluginType := os.Args[2]
+				plugin := os.Args[3]
+
+				switch pluginType {
+				case "provisioner":
```
stbenjam, line 54: This is based on https://github.com/hashicorp/terraform/blob/v0.12.1/command/internal_plugin.go#L48.  `provisioner` is the only thing supported by `internal-plugin` at the moment

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/707/

stbenjam:
CI failure is unrelated, worker didn't seem to come up within the time limit.  I've noticed worker deployments are taking longer lately, maybe we should update the timeout in CI to be a little bit longer? @russellb

hardys:
> CI failure is unrelated, worker didn't seem to come up within the time limit. I've noticed worker deployments are taking longer lately, maybe we should update the timeout in CI to be a little bit longer? @russellb

https://github.com/openshift-metal3/dev-scripts/pull/604 should help with that - I think it would be good to avoid increasing the CI timeout if possible.

hardys:
http://10.8.144.11:8080/job/dev-tools/711/ retriggered CI

data/data/baremetal/masters/main.tf
```diff
@@ -1,3 +1,11 @@
+# This null resource is used to wait for the Ironic API to become available before trying to create
+# the master hosts.
+resource "null_resource" "api-wait" {
+  provisioner "local-exec" {
+    command = "timeout 900 bash -c 'while [[ \"$(curl -s -o /dev/null -w \"%%{http_code}\" ${var.ironic_uri})\" != \"200\" ]]; do sleep 5; done'"
+  }
+}
+
```
hardys, line 8: I wonder if gophercloud would accept a generic function that waits for a specified time until an endpoint responds?

Or there could be a check on https://developer.openstack.org/api-ref/baremetal/#list-api-versions which I think is common to all OpenStack APIs?

This seems fine for a first step though.

hardys: Also there is https://www.terraform.io/docs/providers/http/data_source.html but that doesn't have a timeout option...

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/711/

data/data/baremetal/masters/main.tf
stbenjam: I can't imagine gophercloud taking something like that. I could change the terraform provider to support retrying if this is too ugly, but it felt more natural to me and explicit about what's going on  to handle this using terraform HCL.

stbenjam:
CI failure is current ironic-inspector issue being discussed on slack

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/718/

PR #100, opened 2019-06-10
WIP add Ironic to the bootstrap VM

### LINKED ISSUE #68: Automatically launch Ironic within the bootstrap VM to provision masters
kni-install currently assumes Ironic has been stood up outside of the installer.  Terraform will drive that instance of Ironic to provision the masters.

Instead, we'd like to be able to minimize external pre-requisites and be able to spin up Ironic within the bootstrap VM for provisioning the masters.

Either way, this instance of Ironic is supposed to be temporary.  Once the cluster comes up, the baremetal-operator should be managing Ironic within the cluster.

hardys:
First step to resolve https://github.com/openshift-metal3/kni-installer/issues/68

Currently WIP but feedback welcome

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -7,8 +7,8 @@ IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic-inspe
 IPA_DOWNLOADER_IMAGE=${IPA_DOWNLOADER_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
 COREOS_DOWNLOADER_IMAGE=${COREOS_DOWNLOADER_IMAGE:-"quay.io/openshift-metal3/rhcos-downloader:master"}
 
-# FIXME this should be provided by the installer
-RHCOS_IMAGE_URL="https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/ootpa/410.8.20190520.0"
+# This image is templated in via the installer pkg/asset/ignition/bootstrap/bootstrap.go
+RHCOS_IMAGE_BASE_URL="{{.RHCOSBaseURI}}"
 
 # First we stop any previously started containers, because ExecStop only runs when the ExecStart process
 # e.g this script is still running, but we exit if *any* of the containers exits unexpectedly
@@ -76,7 +76,7 @@ podman run -d --net host --name ipa-downloader \
 
 podman run -d --net host --name coreos-downloader \
      --env CACHEURL="http://172.22.0.1/images" \
-     -v $IRONIC_SHARED_VOLUME:/shared:z ${COREOS_DOWNLOADER_IMAGE} /usr/local/bin/get-resource.sh $RHCOS_IMAGE_URL
+     -v $IRONIC_SHARED_VOLUME:/shared:z ${COREOS_DOWNLOADER_IMAGE} /usr/local/bin/get-resource.sh $RHCOS_IMAGE_BASE_URL
 
 # Start the conductor so db sync can happen while images download
 sudo podman run -d --net host --privileged --name ironic-conductor \
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift-metalkube/kni-installer/pkg/asset/machines"
 	"github.com/openshift-metalkube/kni-installer/pkg/asset/manifests"
 	"github.com/openshift-metalkube/kni-installer/pkg/asset/tls"
+	"github.com/openshift-metalkube/kni-installer/pkg/rhcos"
 	"github.com/openshift-metalkube/kni-installer/pkg/types"
 )
 
@@ -42,6 +43,7 @@ type bootstrapTemplateData struct {
 	PullSecret   string
 	ReleaseImage string
 	Proxy        *configv1.ProxyStatus
+	RHCOSBaseURI string
 }
 
 // Bootstrap is an asset that generates the ignition config for bootstrap nodes.
@@ -208,11 +210,25 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, proxy *c
 		logrus.Debugf("Using internal constant for release image %s", releaseImage)
 	}
 
+	var rhcosImageBaseURI string
+	if ri, ok := os.LookupEnv("OPENSHIFT_INSTALL_IMAGE_BASE_URI_OVERRIDE"); ok && ri != "" {
+		logrus.Warn("Found override for rhcosImageBaseURI. Please be warned, this is not advised")
+		rhcosImageBaseURI = ri
+	} else {
+		var err error
+		rhcosImageBaseURI, err = rhcos.BaseURI()
+		if err != nil {
+			return nil, err
+		}
+		logrus.Debugf("Using internal constant for release image base URI %s", rhcosImageBaseURI)
+	}
+
 	return &bootstrapTemplateData{
 		PullSecret:   installConfig.PullSecret,
 		ReleaseImage: releaseImage,
 		EtcdCluster:  strings.Join(etcdEndpoints, ","),
 		Proxy:        &proxy.Status,
+		RHCOSBaseURI: rhcosImageBaseURI,
 	}, nil
 }
 
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -298,6 +298,7 @@ func (a *Bootstrap) addSystemdUnits(uri string, templateData *bootstrapTemplateD
 		// baremetal platform services
 		"keepalived.service": {},
 		"coredns.service":    {},
+		"ironic.service":     {},
 	}
 
 	directory, err := data.Assets.Open(uri)
--- a/data/data/baremetal/bootstrap/main.tf
+++ b/data/data/baremetal/bootstrap/main.tf
@@ -11,7 +11,7 @@ resource "libvirt_ignition" "bootstrap" {
 resource "libvirt_domain" "bootstrap" {
   name = "${var.cluster_id}-bootstrap"
 
-  memory = "4096"
+  memory = "6144"
 
   vcpu = "4"
 
--- a/data/data/baremetal/main.tf
+++ b/data/data/baremetal/main.tf
@@ -5,6 +5,7 @@ provider "libvirt" {
 provider "ironic" {
   url          = var.ironic_uri
   microversion = "1.52"
+  timeout      = 600
 }
 
 module "bootstrap" {
--- a/data/data/baremetal/main.tf
+++ b/data/data/baremetal/main.tf
@@ -3,7 +3,7 @@ provider "libvirt" {
 }
 
 provider "ironic" {
-  url          = var.ironic_uri
+  url          = "http://${var.bootstrap_provisioning_ip}:6385/v1"
   microversion = "1.52"
   timeout      = 600
 }
--- a/data/data/baremetal/variables-baremetal.tf
+++ b/data/data/baremetal/variables-baremetal.tf
@@ -1,6 +1,6 @@
-variable "ironic_uri" {
+variable "bootstrap_provisioning_ip" {
   type        = string
-  description = "ironic connection URI"
+  description = "IP for the bootstrap VM provisioning nic"
 }
 
 variable "libvirt_uri" {
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/coredns.sh
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/coredns.sh
@@ -25,15 +25,18 @@ if ! podman inspect "$COREDNS_IMAGE" &>/dev/null; then
     podman pull "$COREDNS_IMAGE"
 fi
 MATCHES="$(sudo podman ps -a --format "{{.Names}}" | awk '/coredns$/ {print $0}')"
-if [[ -z "$MATCHES" ]]; then
-    /usr/bin/podman create \
-        --name coredns \
-        --volume /etc/coredns:/etc/coredns:z \
-        --network host \
-        --env CLUSTER_DOMAIN="$CLUSTER_DOMAIN" \
-        --env CLUSTER_NAME="$CLUSTER_NAME" \
-        --env NUM_DNS_MEMBERS="$NUM_DNS_MEMBERS" \
-        --env API_VIP="$API_VIP" \
-        "${COREDNS_IMAGE}" \
-            --conf /etc/coredns/Corefile
+if [[ ! -z "$MATCHES" ]]; then
+    # Remove old pod, or you can get storage for container removed
+    # errors on restart if the container exits unexpectedly
+    podman rm -f coredns
 fi
+/usr/bin/podman create \
+    --name coredns \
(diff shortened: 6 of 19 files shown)


pkg/terraform/exec/plugins/Gopkg.toml
```diff
@@ -66,4 +66,12 @@ ignored = [
 
 [[constraint]]
   name = "github.com/openshift-metalkube/terraform-provider-ironic"
+<<<<<<< HEAD
   version = "v0.1.2"
+=======
+  version = "v0.1.3"
```
stbenjam: Looks like this conflict needs resolving still

data/data/baremetal/masters/main.tf
```diff
@@ -1,3 +1,11 @@
+# This null resource is used to wait for the Ironic API to become available before trying to create
+# the master hosts.
+resource "null_resource" "api-wait" {
+  provisioner "local-exec" {
+    command = "timeout 900 bash -c 'while [[ \"$(curl -s -o /dev/null -w \"%%{http_code}\" ${var.ironic_uri})\" != \"200\" ]]; do sleep 5; done'"
```
stbenjam: I was hoping to have a different way to do this, I tried some things on Friday, but it's tricky. Terraform needs the provider to do the planning stage, so I can't actually block in the provider.  I could do the waiting for each individual resource but that looked worse than this hack.  

This is a known issue: https://github.com/hashicorp/terraform/issues/2430, it's a common problem that "Provider X is responsible for bringing up the infrastructure that provider Y uses."

data/data/bootstrap/files/usr/local/bin/ironic.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
stbenjam: Is the plan for these to become baremetal platform-specific?

stbenjam:
Is there a dev-scripts change to try yet that points to use this Ironic?

data/data/baremetal/masters/main.tf
hardys: Ack, I've been testing this locally and it seems to work fine, perhaps we should go with the curl for now, and be ready to rework it in future?

As discussed yesterday we also need a way to wait for the images, which could be similar e.g `curl -I http://127.0.0.1/images/rhcos-ootpa-latest.qcow2.md5sum` and similar for the IPA kernel/ramdisk?

data/data/bootstrap/files/usr/local/bin/ironic.sh
hardys: Yes, I'll rebase this on https://github.com/openshift-metal3/kni-installer/pull/102

hardys:
> Is there a dev-scripts change to try yet that points to use this Ironic?

I've got a hacked up local branch based on https://github.com/openshift-metal3/dev-scripts/pull/564 - I'll clean that up and push it for testing - I've been working through some issues with podman (buggy errors unmounting the shared volume on pod delete) but will rebase this today.

hardys:
Ok this is getting closer - I had to stop using a pod because the podman/libpod version in the current RHCOS build appears buggy, and fails to unmount the shared volume on pod rm -f, probably a race or something as it works fine with individual containers.

Still TODO is to inject the static NetworkManager configuration for the second nic, I think we can default this to 172.22.0.2 since the default DHCP range for the ironic dnsmasq script doesn't start until 172.22.0.10

I'm also planning to add logic to the *downloader containers to they try to download from a mirror accessible on the default gateway IP (e.g 192.168.111.1) which should mean we can run the downloader containers and httpd on the provisioning host as a cache (at least for development).

@stbenjam I pushed my WIP dev-scripts branch to https://github.com/hardys/dev-scripts/tree/bootstrap_ironic2 - still needs additional work before I open a PR but it's sufficient for my local testing.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -378,6 +379,8 @@ func readFile(name string, reader io.Reader, templateData interface{}) (finalNam
 		if err != nil {
 			return name, data, err
 		}
+		logrus.Tracef("SHDEBUG Rendering file %s", name)
+		logrus.Tracef("SHDEBUG templateData %s", templateData)
```
hardys: This is debugging where I'm trying to work out if the release url is available for templating, doesn't seem to be but perhaps we can add it so the script can be templated with the correct RHCOS_IMAGE_URL

hardys:
Oh and I need to set PROVISIONING_INTERFACE when starting the ironic container, or ensure the configured interface uses the expected name

hardys:
Ok the provisioning nic is now started by startironic using nmcli, and I'm passing the interface name to the container - that exposed that we don't bind only to that interface ref https://github.com/metal3-io/ironic-image/pull/56

I also switched the default Ironic endpoint - I guess if we make Ironic accessible via the public interface of the bootstrap VM we can potentially remove that completely, but for now I've pointed directly at the 172 address used in startironic.sh

hardys:
Ok another problem, ExecStart is supposed to block, but atm the script exits - coredns/keepalived work around this by creating the containers in ExecStartPre, but podman pod start won't work here due to bugs which break removal of the pod.

hardys:
This is now working but I've seen the mariadb container die a couple of times, not exactly clear why yet but I'm wondering if the VM is overloaded so will re-try next week with more memory/CPU to see if that helps

hardys:
Ugh, even after increasing the RAM I'm seeing pods randomly die, and `podman logs` then says no such file or directory from overlayfs - all of this works fine on CentOS with a newer libpod version, so I'm suspecting it's the old 1.0.2-dev podman stuff causing this, but not yet figured out how to rebuild the coreos image with a newer libpod.

hardys:
Ok the latest rebases seem to have fixed the issues I was seeing (although strangely libpod still hasn't been bumped) - this version now works for me locally.

To get this ready to merge I need to resolve the FIXME in startironic where the release version is hard-coded, that should be possible to template in bootstrap.go, although I don't think the release version is currently in the template data.

The other TODO item is to add a survey input for the provisioning network CIDR - we'll need to allow some way to override the default in the case where the subnet is already routable in the local environment.

hardys:
Urgh, spoke too soon, still hitting libpod issues, raised https://bugzilla.redhat.com/show_bug.cgi?id=1724660

stbenjam:
On the next rebase (currently blocked until https://github.com/operator-framework/operator-sdk/pull/1512 is merged), we'll get rhocs 420.8.20190624.0 - hopefully that has a newer libpod.

hardys:
> On the next rebase (currently blocked until [operator-framework/operator-sdk#1512](https://github.com/operator-framework/operator-sdk/pull/1512) is merged), we'll get rhocs 420.8.20190624.0 - hopefully that has a newer libpod.

My understanding is that we need a manual sync which will be tracked via the bz, hopefully that will happen soon then we can rebase on a new build that contains it.

In the meantime I'm going to try rebuilding with coreos-assembler and installing the same (working) version I tested today via rpm-ostree override replace

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -0,0 +1,75 @@
...
+
+podman run -d --net host --name coreos-downloader \
+     --env CACHEURL="http://172.22.0.1/images" \
+     -v $IRONIC_SHARED_VOLUME:/shared:z ${COREOS_DOWNLOADER_IMAGE} /usr/local/bin/get-resource.sh $RHCOS_IMAGE_URL
+
+podman run -d --net host --privileged --name ironic-inspector \
+     --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
+     -v $IRONIC_SHARED_VOLUME:/shared:z "${IRONIC_INSPECTOR_IMAGE}"
+
+podman run -d --net host --privileged --name ironic \
```
dtantsur: Let's split this the same way we did it in dev-scripts: https://github.com/openshift-metal3/dev-scripts/pull/646

dtantsur (changes requested):

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -208,11 +212,25 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, proxy *c
...
+	// baremetal requires some additional templateData to download the RHCOS image
+	// so we can deploy the masters with the correct image from rhcos.json
+	if installConfig.Platform.Name() == bmtypes.Name {
+		var err error
+		deployImage, err = rhcos.Baremetal()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		deployImage = releaseImage
```
markmc: I'd prefer to pass an empty string than confuse the meaning of deployImage - i.e. if it is set, it should be an rhcos image URL, not a release image pullspec

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -208,11 +212,25 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, proxy *c
 		logrus.Debugf("Using internal constant for release image %s", releaseImage)
 	}
 
+	var deployImage string
+	// baremetal requires some additional templateData to download the RHCOS image
+	// so we can deploy the masters with the correct image from rhcos.json
+	if installConfig.Platform.Name() == bmtypes.Name {
+		var err error
+		deployImage, err = rhcos.Baremetal()
```
markmc: something like

```
rhcosImageBaseURI = rhcos.BaseURI()
```

would be more clear

pkg/rhcos/baremetal.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+// Baremetal fetches the BaseURI where baremetal images can be downloaded from
+func Baremetal() (string, error) {
```
markmc: There's nothing specific to bare metal about this - I'd just call it rhcos.BaseURI()

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -0,0 +1,91 @@
...
+
+set -ex
+
+IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
+IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic-inspector:master"}
+IPA_DOWNLOADER_IMAGE=${IPA_DOWNLOADER_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
+COREOS_DOWNLOADER_IMAGE=${COREOS_DOWNLOADER_IMAGE:-"quay.io/openshift-metal3/rhcos-downloader:master"}
+
+# This image is templated in via the installer pkg/asset/ignition/bootstrap/bootstrap.go
+RHCOS_IMAGE_URL="{{.DeployImage}}"
```
markmc: Make it `RHCOS_IMAGE_BASE_URL` or similar

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -42,6 +44,7 @@ type bootstrapTemplateData struct {
 	PullSecret   string
 	ReleaseImage string
 	Proxy        *configv1.ProxyStatus
+	DeployImage  string
```
markmc: Again, `RHCOSBaseURI` or similar would be more clear

You could leave a comment here that the baremetal platform is the only user currently

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -208,11 +212,25 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, proxy *c
 		logrus.Debugf("Using internal constant for release image %s", releaseImage)
 	}
 
+	var deployImage string
+	// baremetal requires some additional templateData to download the RHCOS image
+	// so we can deploy the masters with the correct image from rhcos.json
+	if installConfig.Platform.Name() == bmtypes.Name {
```
markmc: No need for this to be bare metal specific - just make baseURI available for templating by all platforms

markmc (changes requested):
Just focused on the baseURI templating in this review - I suggest splitting that out into a separate commit, and maybe even proposing it to openshift/installer upstream now

pkg/asset/ignition/bootstrap/bootstrap.go
hardys: Thanks for the feedback @markmc this is all now working locally for me so I kicked off a CI run via https://github.com/openshift-metal3/dev-scripts/pull/642 to replicate that, but in the meantime I will rework this to address your comments, and when you're happy we can break out that commit and propose upstream (unless @stbenjam plans to push a PR with this combination of commits soon, I don't want to complicate that..)

hardys:
@markmc https://github.com/openshift-metal3/kni-installer/pull/100/commits/29bd776c0d37ed9f046996d0ec68742ab249cf42 should address your comments I think, please re-review when you get a moment.

@stbenjam Note the latest commit added which swaps out IronicURI for BootstrapProvisioningIP - if you're happy with that approach I'll use the data to template the IP for startironic.sh, and subsequently we can add similar interfaces to customize the DHCP pool for provisioning?

The alternative would be to pass a CIDR, but then we'll have hard-coded assumptions about which IP in the CIDR is used for each thing, I was preferring to just expose the things which need IPs separately but open to feedback on the approach.

hardys:
Hmm, testing indicates I broke something when renaming as startironic.sh gets an empty string for `{{.RHCOSBaseURI}}`

stbenjam:
> @stbenjam Note the latest commit added which swaps out IronicURI for BootstrapProvisioningIP - if you're happy with that approach I'll use the data to template the IP for startironic.sh, and subsequently we can add similar interfaces to customize the DHCP pool for provisioning?
> 
> The alternative would be to pass a CIDR, but then we'll have hard-coded assumptions about which IP in the CIDR is used for each thing, I was preferring to just expose the things which need IPs separately but open to feedback on the approach.

I think this approach is good for now, but I'd prefer a CIDR eventually and let us pick the IP because we'll want to assign static IP's on the provisioning network to the masters for install-gather debugging. But there's probably all kinds of things to figure out there that it should probably be it's own task so we can deal with it later.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -208,11 +210,25 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, proxy *c
 		logrus.Debugf("Using internal constant for release image %s", releaseImage)
 	}
 
+	var rhcosImageBaseURI string
+	if ri, ok := os.LookupEnv("OPENSHIFT_INSTALL_IMAGE_BASE_URI_OVERRIDE"); ok && ri != "" {
```
hardys, line 214: @derekhiggins I think it would perhaps be better if we used this override instead of CACHEURL, because it's one less thing we'll need to hard-code (and ultimately template) in startironic.sh - what are your thoughts on instead making the https://github.com/openshift-metal3/rhcos-downloader/blob/master/get-resource.sh output similar to the original repo structure, or we can cache differently in dev-scripts, such that we just set OPENSHIFT_INSTALL_IMAGE_BASE_URI_OVERRIDE in dev-scripts?

hardys: Cancel that, @abhinavdahiya has indicated in https://github.com/openshift/installer/pull/2035 that we shouldn't add a new variable like this

russellb:
Since the kni-installer repo is no longer in use and we're now tracking this via https://github.com/openshift/installer/issues/2060, I'm going to close this PR.  I know you'll be posting this against openshift/installer when you're ready

Issue #64, opened 2019-04-30
Complete addition of VIP management to installer


russellb:
VIP management using keepalived is partially integrated into the installer, and also via dev-scripts.  The remaining pieces from dev-scripts need to be moved into the installer.  This should also be conditional based on platform type.

hardys:
I think #102 closes this based on the review comments from @cybertron

PR #102, opened 2019-06-11
Enable platform dependent bootstrap ignition files

### LINKED ISSUE #62: Complete addition of coredns-mdns integration to installer
Our integration of coredns-mdns is not specific to the `baremetal` platform.  Before this can be upstreamed to OpenShift, it should be updated to only be run on the platform(s) where it's needed.

hardys:
Some of the additions made should only be enabled for the
baremetal platform, so move these to data/data/bootstrap/baremetal
and adjust the code to conditionally read this depending on the
platform name.

Closes: #62
Related: #64


hardys:
Only lightly tested so far, but lets ensure it passes CI, feedback welcome on the approach - I discussed it with @staebler and it sounds like it may be a reasonable approach to enable platform specific additions.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -127,6 +127,24 @@ func (a *Bootstrap) Generate(dependencies asset.Parents) error {
...
+	// Check for optional platform specific directories
+	platform := installConfig.Config.Platform.Name()
+	platformFilePath := fmt.Sprintf("bootstrap/%s/files", platform)
+	if _, err := os.Stat(platformFilePath); !os.IsNotExist(err) {
+		err = a.addStorageFiles("/", platformFilePath, templateData)
+		if err != nil {
+			return err
+		}
+	}
+	platformUnitPath := fmt.Sprintf("bootstrap/%s/units", platform)
```
staebler: This should be `bootstrap/%s/systemd/units`.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -127,6 +127,24 @@ func (a *Bootstrap) Generate(dependencies asset.Parents) error {
 	if err != nil {
 		return err
 	}
+
+	// Check for optional platform specific directories
+	platform := installConfig.Config.Platform.Name()
+	platformFilePath := fmt.Sprintf("bootstrap/%s/files", platform)
+	if _, err := os.Stat(platformFilePath); !os.IsNotExist(err) {
```
staebler: This check does not honor the `OPENSHIFT_INSTALL_DATA` environment variable.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -127,6 +127,24 @@ func (a *Bootstrap) Generate(dependencies asset.Parents) error {
...
+		err = a.addStorageFiles("/", platformFilePath, templateData)
+		if err != nil {
+			return err
+		}
+	}
+	platformUnitPath := fmt.Sprintf("bootstrap/%s/units", platform)
+	if _, err := os.Stat(platformUnitPath); !os.IsNotExist(err) {
```
staebler: Same concern as above regarding the `OPENSHIFT_INSTALL_DATA` environment variable.

staebler (commented):
The approach taken elsewhere is to place files that should be used on all platform in a `_base` folder. For example, bootkube.service would move to `data/data/bootstrap/_base/systemd/units/bootkube.service`.

Also, there is no `baremetal` platform type in the installer. Bare metal installations use the `none` platform. But note that not all `none`-platform installations are bare-metal installations.

stbenjam:
@staebler There is a baremetal platform in our fork (it's why it exists), it will be contributed back to openshift/installer soon. This fork implements baremetal IPI.

stbenjam:
@hardys I think the approach is good.  The only concern I have is that looking at the implementation of addStorageFiles, I don't think it would be possible for a platform to override a base configuration file.  If for example, we wanted a custom `motd`, I believe the same file would end up in the ignition twice - not sure how ignition handles it, but ideally we'd create a unique set of files to deploy, with the platform-specific ones taking precedence

staebler:
> @staebler There is a baremetal platform in our fork (it's why it exists), it will be contributed back to openshift/installer soon. This fork implements baremetal IPI.

Ah, I didn't even notice that this wasn't https://github.com/openshift/installer.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/731/

pkg/asset/ignition/bootstrap/bootstrap.go
hardys: Oops, thanks!

hardys:
> The approach taken elsewhere is to place files that should be used on all platform in a `_base` folder. For example, bootkube.service would move to `data/data/bootstrap/_base/systemd/units/bootkube.service`

Thanks for the feedback - for this fork I wanted to avoid causing us more rebase pain, so decided not to move the files, but it sounds like I should create another PR to openshift/installer which aligns with that convention, then we can rebase and drop this patch when it hopefully lands :)

As mentioned this fork implements a new baremetal platform, and this is part of our cleanup to enable proposing the additions to enable baremetal IPI without impact to other platforms.

hardys:
> @hardys I think the approach is good. The only concern I have is that looking at the implementation of addStorageFiles, I don't think it would be possible for a platform to override a base configuration file. If for example, we wanted a custom `motd`, I believe the same file would end up in the ignition twice - not sure how ignition handles it, but ideally we'd create a unique set of files to deploy, with the platform-specific ones taking precedence

This is a good point, I need to do some testing but AFAICS recent ignition versions do validation so probably duplicate files will break, so we'll need to handle it in the installer.

I'll look into that, but I think the other issues raised by @staebler are now fixed.

derekhiggins:
Build ABORTED, see build http://10.8.144.11:8080/job/dev-tools/737/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/738/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/739/

pkg/asset/ignition/bootstrap/bootstrap.go
hardys: Thanks, I switched to using data.Assets in the latest revision which I think should resolve this

pkg/asset/ignition/bootstrap/bootstrap.go
hardys: Thanks, I switched to using data.Assets in the latest revision which I think should resolve this

hardys:
@stbenjam I took a look at how we can handle colliding files/units, and I think we'll need to add some logic to track the append index for each filename, then insert instead of append when we find a colliding name.

Since that's logically separate from this change, and we don't yet have any colliding files, perhaps that can be a followup if/when needed?

stbenjam:
Sure, sounds good to me.

I also agree with deferring the move to `_base` - it'd be very painful to rebase if any of those files change.

stbenjam (approved):

hardys:
@celebdor, @yboaron  @cybertron can you check you're happy with this as it touches the keepalived and coredns integration, also please can you confirm if this will close both #62 and #64 ?

If you're happy then I think we're probably good to merge this.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -247,10 +269,11 @@ func (a *Bootstrap) addSystemdUnits(uri string, templateData *bootstrapTemplateD
 		"progress.service":                {},
 		"kubelet.service":                 {},
 		"chown-gatewayd-key.service":      {},
-		"keepalived.service":              {},
-		"coredns.service":                 {},
 		"systemd-journal-gatewayd.socket": {},
 		"approve-csr.service":             {},
+		// baremetal platform services
```
cybertron, line 274: Long term our plan is to move away from using systemd services on the bootstrap node too, but in the meantime this looks good.

cybertron (approved):
Looks like this does wrap up the installer parts of those changes. The behavior here mirrors what happens in MCO for platform-specific files, so lgtm.

PR #103, opened 2019-06-13
Add ingress_vip to installer survey

cybertron:
We want to get away from looking up vips at deploy time from DNS.
Instead, they should be specified in install-config.yaml and looked
up from that data.

This adds ingress_vip as a configuration option like the previously
added api_vip.

### DIFF
--- a/pkg/asset/installconfig/baremetal/baremetal.go
+++ b/pkg/asset/installconfig/baremetal/baremetal.go
@@ -12,7 +12,7 @@ import (
 
 // Platform collects bare metal specific configuration.
 func Platform() (*baremetal.Platform, error) {
-	var libvirtURI, ironicURI, externalBridge, provisioningBridge, apiVIP string
+	var libvirtURI, ironicURI, externalBridge, provisioningBridge, apiVIP, ingressVIP string
 	var hosts []*baremetal.Host
 
 	err := survey.Ask([]*survey.Question{
@@ -78,9 +78,23 @@ func Platform() (*baremetal.Platform, error) {
 				Help:    "The VIP to be used for internal API communication. If blank, the address will be looked up from DNS.",
 				Default: baremetaldefaults.APIVIP,
 			},
-			Validate: ipValidator,
+			Validate: apiVIPValidator,
 		},
 	}, &apiVIP)
+	if err != nil {
+		return nil, err
+	}
+
+	err = survey.Ask([]*survey.Question{
+		{
+			Prompt: &survey.Input{
+				Message: "Ingress VIP",
+				Help:    "The VIP to be used for ingress. If blank, the address will be looked up from DNS.",
+				Default: baremetaldefaults.IngressVIP,
+			},
+			Validate: ingressVIPValidator,
+		},
+	}, &ingressVIP)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/types/baremetal/defaults/platform.go
```diff
@@ -51,4 +52,15 @@ func SetPlatformDefaults(p *baremetal.Platform, c *types.InstallConfig) {
 			p.ApiVIP = vip[0]
 		}
 	}
+
+	if p.IngressVIP == IngressVIP {
+		// This name should resolve to exactly one address
+		vip, err := net.LookupHost("test.apps." + c.ClusterDomain())
```
cybertron, line 59: Is this a safe assumption for the ingress DNS entry?

hardys:
Sorry @cybertron looks like this now needs a rebase, we landed various fixes related to getting the first openshift-install PR up over the last few days.

pkg/types/baremetal/defaults/platform.go
hardys: @celebdor or @yboaron can you comment on this - I think this should work and always resolve to the ingress VIP, right?

yboaron: @shardy - that's correct.
The Ingress VIP should be defined as wildcard dns record (*.apps.$ClusterDomain),
so any prefix added to  '.apps.$ClusterDomain' should resolve to the ingress VIP.

pkg/asset/installconfig/baremetal/baremetal.go
```diff
@@ -123,8 +138,20 @@ func uriValidator(ans interface{}) error {
 	return validate.URI(ans.(string))
 }
 
+<<<<<<< HEAD
```
cybertron: Whoops, missed a merge conflict. :-(

pkg/types/baremetal/defaults/platform.go
cybertron: I guess my main concern was whether the .apps. part was always the case. Is the prefix hard-coded or can users configure that?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/811/

pkg/types/baremetal/defaults/platform.go
yboaron: @cyberton,  Hmm, that's a good point, I was 100% sure that it's hardcoded value, so according to OCP4.1 documentation [1], seems that it should be hardcoded value for baremetal UPI.  

To double check that I sent a Q to it OCP slack channel [2]

[1] https://docs.openshift.com/container-platform/4.1/installing/installing_bare_metal/installing-bare-metal.html#installation-dns-user-infra_installing-bare-metal 
[2] https://coreos.slack.com/archives/C68TNFWA2/p1561533033492300

yboaron: @cybertron , it should be safe to use '.apps.$ClusterDomain', see [1]
[1] https://coreos.slack.com/messages/CCH60A77E/convo/CCH60A77E-1561545440.277400/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/826/

Issue #58, opened 2019-04-24
Consider dropping deploy kernel/ramdisk from install config


russellb:
The install config currently includes options like:

```
          deploy_kernel:  "http://172.22.0.1/images/ironic-python-agent.kernel"
          deploy_ramdisk: "http://172.22.0.1/images/ironic-python-agent.initramfs"
```

We should aim for Ironic to be an implementation detail, and this leaks a bit of info.  We also don't expect this image to ever change.  Since we know exactly what image should be used, the installer should automatically download it and ensure it's available via HTTP from somewhere.

stbenjam:
Should the installer do it? Or could we have the Ironic container take care of it? If the installer needs to do it, where do we host it? We'd have to have it available from somewhere in the bootstrap environment, and also later in the running cluster.

russellb:
Yeah, that's a good idea.  Maybe we could include it in our ironic container?

I'm not tied to any particular solution.  I just think making this not a concern of the install-config should be a todo item.

dhellmann:
When we talked about this in the context of packaging the operator and actuator, we tossed around the idea of having the operator provide a web server container with these things and the actuator providing a web server containing the images to be deployed. @imain was looking at that.

russellb:
That works for day 2, but not the day 1 installer when the operator and actuator aren't running yet.

dhellmann:
Good point. We also talked about that container just being a cache, so maybe we can use the same container image in both cases?

hardys:
I think with the new downloader containers, and the work to move those and the ironic services to the bootstrap VM, we can make progress on this soon.

Ref https://github.com/openshift-metal3/kni-installer/pull/100 - the IP of the bootstrap VM will be known (either hardcoded or available via a survey input), so we will know where the ramdisk/kernel and RHCOS image should be when deploying via terraform (the same approach can be used on cluster too, but the URL will obviously be different).

One question is what do we put in the baremetal host definition for the masters, when the URL for the images goes away as soon as bootstrap is completed?

dhellmann:
> One question is what do we put in the baremetal host definition for the masters, when the URL for the images goes away as soon as bootstrap is completed?

The master hosts will be registered with a machine reference but no image information. That will make them appear as "externally provisioned" and prevent the operator from trying to re-provision them when it is launched inside the cluster.

hardys:
[...]

PR #104, opened 2019-06-13
[WIP] Install gather support for baremetal platform

### LINKED ISSUE #79: Add baremetal support for install gather
openshift added debugging - https://github.com/openshift/installer/blob/master/cmd/openshift-install/gather.go - on installation failure.  We need to add support for the baremetal platform.  In order to do that, terraform needs to know the IP's of the bootstrap and master nodes.

This produces some confusing error messages:

```
level=debug msg="Fetching \"Install Config\"..."
level=debug msg="Loading \"Install Config\"..."
level=debug msg="  Loading \"SSH Key\"..."
level=debug msg="  Loading \"Base Domain\"..."
level=debug msg="    Loading \"Platform\"..."
level=debug msg="  Loading \"Cluster Name\"..."
level=debug msg="    Loading \"Base Domain\"..."
level=debug msg="  Loading \"Pull Secret\"..."
level=debug msg="  Loading \"Platform\"..."
level=debug msg="Using \"Install Config\" loaded from state file"
level=debug msg="Reusing previously-fetched \"Install Config\""
level=error msg="Cannot fetch the bootstrap and control plane host addresses from state file for baremetal platform"
level=error msg="boostrap host address and at least one control plane host address must be provided"
```

This isn't the actual failure, it's just a failure to gather debugging data. The real problem should be elsewhere in the log (usually a timeout)

stbenjam:
fixes #79

This adds support to the 'gather' command for baremetal. The bootstrap
node's IP address is retrieved via libvirt's DHCP records.  The master's
IP's are retrieved using Ironic introspection data, that is stored in
the terraform state's introspection data source (added in
terraform-provider-ironic v0.1.4)

How to test this:

- Kick off an install with this PR checked out, ensuring you set `KNI_INSTALL_FROM_GIT=true` env variable
- When the installer reaches the point where it's waiting for the bootstrap API to come up, hit control + C to abort
- Run the gather command:

```console
[root@dell-r730-021 dev-scripts]# ~/go/src/github.com/openshift-metalkube/kni-installer/bin/kni-install gather bootstrap --dir ocp
INFO Use the following commands to gather logs from the cluster 
INFO ssh -A core@192.168.111.38 '/usr/local/bin/installer-gather.sh 172.22.0.97 172.22.0.10 172.22.0.14' 
INFO scp core@192.168.111.38:~/log-bundle.tar.gz . 
```


pkg/terraform/gather/baremetal/ip.go
```diff
@@ -0,0 +1,108 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"strings"
+
+	"github.com/openshift-metalkube/kni-installer/pkg/terraform"
+)
+
+// BootstrapIP returns the ip address for bootstrap host. Baremetal relies on a libvirt bootstrap node, that gets it's
+// IP from DHCP.
+func BootstrapIP(tfs *terraform.State, config *types.InstallConfig) (string, error) {
```
stbenjam, line 19: This is all bit complex, since we have to retrieve the lease information from libvirt.  The libvirt platform generates the IP's: https://github.com/openshift/installer/blob/master/pkg/tfvars/libvirt/libvirt.go, maybe we could use that approach instead.

We're also affected by https://github.com/openshift-metal3/dev-scripts/issues/512 here

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/746/

pkg/terraform/gather/baremetal/ip.go
hardys: What do we need the BootstrapIP for?  I'm wondering if we can use the static IP configured for the provisioning network in https://github.com/openshift-metal3/kni-installer/pull/100 instead?

stbenjam: Debug gathering gets logs from bootstrap if install fails while bootstrap is still up, I think it also uses it as a bastion host to get to the masters in that case as well. 

But we can definitely use the static IP instead of this.

hardys: I've run into issues getting ironic to work reliably on the boostrap VM, but I may try breaking #100 into two or three PRs, then we can enable the provisioning nic so that this PR can use it, then potentially host the images to allow fixing https://github.com/openshift-metal3/kni-installer/issues/58, and finally switch the ironic endpoint over to the bootstrap VM.

@stbenjam let me know if the above decoupling of PRs would be helpful, if so I'll rebase and open some additional PRs.

stbenjam: I think that makes sense. I don't think we need install gather to open the first PR to openshift/installer but we do need it eventually.  

Could we move the masters to static IP's as well? Or do you think the approach of getting them out of ironic introspection data is reliable?

hardys: Yes I think we'll need to switch the masters to static IPs to enable self-hosted ironic where we'll need to run dnsmasq and the other ironic bits on a master to deploy the workers - ref discussion on https://github.com/metal3-io/metal3-docs/pull/39

I'm not yet sure what the best way to inject that configuration is though, e.g we could use ignition, something run via the BMO or perhaps the MCO, and there's the nmstate operator to investigate too.

stbenjam:
Marking this WIP again, as it sounds like we can just use static IP's

stbenjam:
I'm going to close this, we can open a PR to enable gather when we're using static ip's.

hardys:
> I'm going to close this, we can open a PR to enable gather when we're using static ip's.

Note that https://github.com/openshift/installer/pull/1941 landed yesterday, I'm testing locally with the new RHCOS version and plan to update https://github.com/openshift-metal3/kni-installer/pull/100 soon, so that should hopefully provide the static provisioning IP needed for the bootstrap VM.

stbenjam:
> > I'm going to close this, we can open a PR to enable gather when we're using static ip's.
> 
> Note that [openshift/installer#1941](https://github.com/openshift/installer/pull/1941) landed yesterday, I'm testing locally with the new RHCOS version and plan to update #100 soon, so that should hopefully provide the static provisioning IP needed for the bootstrap VM.

We're going to move to static IP's for masters on the provisioning network as well, aren't we?  Or do we let DHCP handle it and just get the data from introspection for install-gather?

PR #105, opened 2019-06-18
WIP: pkg/asset/manifests/infrastructure: Set status.platformStatus.ba…

bcrochet:
…remetal.apivip

Using openshift/api@6a276bd3 (config/v1/types_infrastructure: Add BareMetal
status for BareMetal platform, 2019-06-11, openshift/api#348) so static
pods can use the API VIP in their configuration.


russellb:
This will need to be re-posted against openshift/installer

russellb:
This is now tracked here: https://github.com/openshift/installer/issues/2066

Please re-post against openshift/installer when you're ready

PR #89, opened 2019-05-28
Add initial baremetal IPI docs

stbenjam:
fixes #67

In order to get an experimental version of the baremetal platform in openshift/installer, we need to have some basic docs about how to use it.

The documentation about our fork, while it exists, can be moved into a wiki on this repo.  This PR restores the openshift readme, and adds some basic docs about the platform.

PR #109, opened 2019-06-19
Restore documentation to openshift latest

stbenjam:
I've moved kni-install specific info to the [Wiki](https://github.com/openshift-metal3/kni-installer/wiki). This
restores the docs to the upstream, so we're not including fork-specific info in the openshift/installer PR.

I have another PR open to add some basic docs about how to use the baremetal IPI platform with the installer, but that isn't ready yet - https://github.com/openshift-metal3/kni-installer/pull/89.


hardys:
Sounds reasonable - if for any reason the wiki proves inconvenient we could move the docs to the openshift-metal3-docs repo instead

stbenjam:
I am hoping this entire repo doesn't last very long :-) :crossed_fingers:

PR #112, opened 2019-06-19
Remove vendoring of openshift/installer

stbenjam:
We somehow vendored openshift/installer, I imagine due to something with
the rename. There's no trace of it in vendor/, but it ended up in the
lock file.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/770/

PR #115, opened 2019-06-19
Add build flags for baremetal destroy

stbenjam:
Otherwise building installer without baremetal fails with errors, e.g.
when just building an installer for openstack you end up with:

```
pkg/destroy/baremetal/baremetal.go:24:12: undefined: "github.com/openshift/installer/vendor/github.com/libvirt/libvirt-go".NewConnect
```

libvirt also conditionally builds their destroy package for the same
reasons.

### DIFF
--- a/pkg/destroy/baremetal/baremetal.go
+++ b/pkg/destroy/baremetal/baremetal.go
@@ -1,3 +1,5 @@
+// +build ironic
+
 package baremetal
 
 import (
--- a/pkg/destroy/baremetal/register.go
+++ b/pkg/destroy/baremetal/register.go
@@ -1,3 +1,5 @@
+// +build ironic
+
 // Package baremetal provides a cluster-destroyer for bare metal clusters.
 package baremetal
 


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/772/

pkg/destroy/baremetal/baremetal.go
```diff
@@ -1,3 +1,5 @@
+// +build ironic
```
hardys, line 1: I'm missing something here - why is the build tag "ironic" and not "baremetal"?

stbenjam: It probably should be baremetal, but the tag we use is ironic.

hardys: Ah I see it's already used for some of the terraform bits - it seems weird to expose an internal implementation detail to those building, and it's inconsistent with other platforms, so perhaps we can do a follow-up which fixes all the tags to match the platform name?

PR #126, opened 2019-06-21
WIP: Pass keepalived image to mco bootstrap

cybertron:
I believe we will need to do this once we are deploying our services
with mco instead of dev-scripts.


stbenjam:
Is this based on the comment here: https://github.com/openshift/machine-config-operator/pull/795#issuecomment-504567632?

cybertron:
No, this is just my attempt to get us using the images from the release instead of pointing at our external repos on quay.io.

markmc:
Principle seems sane - for any images we use, we need to get its pullspec from the release image

bcrochet (commented):
We will likely need to add the other images too... Coredns, haproxy, mdns-publisher, baremetal-runtimecfg.

russellb:
This is now tracked here: https://github.com/openshift/installer/issues/2067

PR #132, opened 2019-07-02
vendor: match terraform-provider-openstack with openshift/installer

stbenjam:
openshift/installer is now vendoring a version of gophercloud compatible
with the Ironic terraform provider (in fact, the exact sha we vendored). 

When we bumped gophercloud (before they did), I pinned the openstack
provider to a compatible release. They chose a slightly different
version of terraform-provider-openstack when they did theirs.

Let's match the same version they're using.


pkg/terraform/exec/plugins/Gopkg.toml
```diff
@@ -34,7 +34,7 @@ ignored = [
 
 [[constraint]]
   name = "github.com/terraform-providers/terraform-provider-openstack"
-  version = "=1.19.0"
+  revision = "b1406b8e4894faad993aff786f0bb50bfec8e281"
```
stbenjam, line 37: https://github.com/openshift/installer/blob/82830c4d89b0046bd3252e914a7910573c23c344/pkg/terraform/exec/plugins/Gopkg.toml#L37

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/820/

PR #135, opened 2019-07-04
Fix typo in JSON for baremetal platform

stbenjam:
I was looking at the various commits since the last release until now, and I noticed this typo. I do not think this is why things are failing (but neither did we think the last typo in a JSON annotation was all that serious).


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/831/

PR #139, opened 2019-07-08
Add user docs for baremetal IPI.

### LINKED ISSUE #67: Address documentation gaps
https://github.com/openshift-metal3/kni-installer/blob/master/docs/dev/baremetal.md

Our dev doc includes some information that's only relevant to the fork.  To prepare a PR, should move the fork-specific content elsewhere.  Perhaps we could use a wiki page on the kni-installer repo?  Some of the dev doc content makes sense to keep in the repo.

We also need to create a user doc that covers how the platform is used, prerequisites, install config syntax, and so forth.

russellb:
Closes #67

Co-authored-by: Stephen Benjamin <stephen@redhat.com>

### DIFF
--- a/docs/user/metal/README.md
+++ b/docs/user/metal/README.md
@@ -1,0 +1,22 @@
+# Support for Bare Metal Environments
+
+OpenShift has support for bare metal deployments with either [User
+provided infrastructure (UPI)](install_upi.md), or [Installer-provided
+instrastructure (IPI)](install_ipi.md).
+
+The following is a summary of key differences:
+
+* UPI bare metal
+  * Provisioning hosts is an external requirement
+  * Requires extra DNS configuration
+  * Requires setup of load balancers
+  * Offers more control and choice over infrastructure
+
+* IPI bare metal
+  * Has built-in hardware provisioning components, will provision nodes with RHCOS automatically,
+    and supports the Machine API for ongoing management of these hosts.
+  * Automates internal DNS requirements
+  * Automates setup of self-hosted load balancers
+  * Supports “openshift-install create cluster” for bare metal environments
+    using this infrastructure automation, but requires the use of compatible
+    hardware, as described in [install_ipi.md](install_ipi.md).
--- a/docs/user/metal/install_ipi.md
+++ b/docs/user/metal/install_ipi.md
@@ -1,0 +1,225 @@
+# Bare Metal IPI (Installer Provisioned Infrastructure) Overview
+
+Current Status: **Experimental**
+
+This document discusses the installer support for an IPI (Installer Provisioned
+Infrastructure) install for bare metal hosts.  This includes platform support
+for the management of bare metal hosts, as well as some automation of DNS and
+load balancing to bring up the cluster.
+
+The upstream project that provides Kubernetes-native management of bare metal
+hosts is [metal3.io](http://metal3.io).
+
+For UPI (User Provisioned Infrastructure) based instructions for bare metal
+deployments, see [install_upi.md](install_upi.md).
+
+## Prerequisites
+
+### Ironic
+
+Currently, the `baremetal` platform requires an existing Ironic environment.
+This will eventually be handled by `openshift-install`, with Ironic being
+deployed onto the bootstrap node. Until then, users of the `baremetal` platform
+should use the
+[openshift-metal3/dev-scripts](https://github.com/openshift-metal3/dev-scripts)
+repository to handle configuration of Ironic.
+
+The following PR contains the WIP changes for automating Ironic from
+`openshift-install`: https://github.com/openshift-metal3/kni-installer/pull/100
+
+### Network Requirements
+
+It is assumed that all hosts have at least 2 NICs, used for the following
+purposes:
+
+* **NIC #1 - External Network**
+  * This network is the main network used by the cluster, including API traffic
+    and application traffic.
+  * ***DHCP***
+    * External DHCP is assumed on this network.  It is **strongly** recommended
+      to set up DHCP reservations for each of the hosts in the cluster to
+      ensure that they retain stable IP addresses.
+    * A pool of dynamic addresses should also be available on this network, as
+      the provisioning host and temporary bootstrap VM will also need addresses
+      on this network.
+  * ***NTP***
+    * A time source must be accessible from this network.
+  * ***Reserved VIPs (Virtual IPs)*** - 3 IP addresses must be reserved on this
+    network for use by the cluster.  Specifically, these IPs will serve the
+    following purposes:
+    * API - This IP will be used to reach the cluster API.
+    * Ingress - This IP will be used by cluster ingress traffic
+    * DNS - This IP will be used internally by the cluster for automating
+      internal DNS requirements.
+  * ***External DNS*** - While the cluster automates the internal DNS
+    requirements, two external DNS records must be created in whatever DNS
+    server is appropriate for this environment.
+    * `api.<cluster-name>.<base-domain>` - pointing to the API VIP
+    * `*.apps.<cluster-name>.<base-domain>` - pointing to the Ingress VIP
+
+* **NIC #2 - Provisioning Network**
+  * A private, non-routed network, used for PXE based provisioning.
+  * DHCP is automated for this network.
+  * Addressing for this network is currently hard coded as `172.22.0.0/24`, but
+    will be made configurable in the future.
+
+* **Out-of-band Management Network**
+  * Servers will typically have an additional NIC used by the onboard
+    management controllers (BMCs).  These BMCs must be accessible and routed to
+    the host.
+
+### Provisioning Host
+
+The installer must be run from a host that is attached to the same networks as
+the cluster, as described in the previous section.  We refer to this host as
+the *provisioning host*.  The easiest way to provide a provisioning host is to
+use one of the hosts that is intended to later become a worker node in the same
+cluster.  That way it is already connected to the proper networks.
+
+It is recommended that the provisioning host be a bare metal host, as it must be
+able to use libvirt to launch the OpenShift bootstrap VM locally.
+
+### Supported Hardware
+
+The architecture is intended to support a wide variety of hardware.  This was
+one of the reasons Ironic is used as an underlying technology.  However, so far
+development and testing has focused on PXE based provisioning using IPMI for
+out-of-band management of hosts.  Other provisioning approaches will be added,
+tested, and documented over time.
+
+## Installation Process
+
+Once an environment has been prepared according to the documented
+pre-requisites, the install process is the same as other IPI based platforms.
+
+`openshift-install create cluster`
+
+However, it is recommended to prepare an `install-config.yaml` file in advance,
+containing all of the details of the bare metal hosts to be provisioned.
+
+### Install Config
+
+The `install-config.yaml` file requires some additional details.  Most of the
+information is teaching the installer and the resulting cluster enough about
+the available hardware so that it is able to fully manage it.
+
+Here is an example `install-config.yaml` with the required `baremetal` platform
+details.
+
+**IMPORTANT NOTE:** The current install configuration for the `baremetal`
+platform should be considered experimental and still subject to change without
+backwards compatibility.  In particular, some items likely to change soon
+include:
+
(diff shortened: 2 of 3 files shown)


docs/user/metal/README.md
```diff
@@ -0,0 +1,22 @@
+# Support for Bare Metal Environments
+
+OpenShift has support for deployments on baremetal with either [User
```
stbenjam: Per the suggested style guidelines, https://github.com/metal3-io/metal3-docs/pull/47, I think this should be `on bare metal hosts`

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,262 @@
...
+      internal DNS requirements.
+  * ***External DNS*** - While the cluster automates the internal DNS
+    requirements, two external DNS records must be created in whatever DNS
+    server is appropriate for this environment.
+    * `api.<cluster-name>.<base-domain>` - pointing to the API VIP
+    * `*.apps.<cluster-name>.<base-domain>` - pointing to the Ingress VIP
+
+* **NIC #2 - Provisioning Network**
+  * A private, non-routed network, used for PXE based provisioning.
+  * DHCP is automated for this network.
```
stbenjam, line 62: Should we mention this is presently hardcoded, but in the future we will let the user specify a range?

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Provisioning Host
+
+The installer must be run from a host that is attached to the same networks as
+the cluster, as described in the previous section.  We refer to this host as
+the *provisioning host*.  The easiest way to provide a provisioning host is to
+use one of the hosts that is intended to later become a worker node in the same
+cluster.  That way it is already connected to the proper networks.
+
+It is recommended that the provisioning host be a baremetal host, as it must be
```
stbenjam: s/baremetal/bare metal/

docs/user/metal/README.md
russellb: oops.

docs/user/metal/install_ipi.md
russellb: thanks

docs/user/metal/install_ipi.md
russellb: Good idea

stbenjam (approved):
lgtm

docs/user/metal/README.md
```diff
@@ -0,0 +1,22 @@
...
+  * Requires setup of load balancers
+  * Offers more control and choice over infrastructure
+
+* IPI bare metal
+  * Has built in hardware provisioning components, will provision nodes with RHCOS automatically,
+    and supports the Machine API for ongoing management of these hosts.
+  * Automates internal DNS requirements
+  * Automates setup of self-hosted load balancers
+  * Supports “openshift-install create cluster” for bare metal environments
+    using this infrastructure automation, but imposes more restructions on
```
markmc: restrictions

docs/user/metal/README.md
```diff
@@ -0,0 +1,22 @@
...
+The following is a summary of key differences:
+
+* UPI bare metal
+  * Provisioning hosts is an external requirement
+  * Requires extra DNS configuration
+  * Requires setup of load balancers
+  * Offers more control and choice over infrastructure
+
+* IPI bare metal
+  * Has built in hardware provisioning components, will provision nodes with RHCOS automatically,
```
markmc: built-in ?

docs/user/metal/README.md
```diff
@@ -0,0 +1,22 @@
...
+    and supports the Machine API for ongoing management of these hosts.
+  * Automates internal DNS requirements
+  * Automates setup of self-hosted load balancers
+  * Supports “openshift-install create cluster” for bare metal environments
+    using this infrastructure automation, but imposes more restructions on
+    hardware.
```
markmc: That really begs the question "what restrictions?" Probably need to explain a little here or point to where it's explained?

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,264 @@
+# Bare Metal IPI (Installer Provisioned Infrastructure) Overview
+
+Current Status: **Experimental**
+
+This document discusses the installer support for an IPI (Installer Provisioned
+Infrastructure) install for bare metal hosts.  This includes platform support
+for the management of bare metal hosts, as well as some automation of DNS and
+load balancing to bring up the cluster.
+
+The upstream project that provides Kubernetes native management of bare metal
```
markmc: Kubernetes-native

So says steveg :)

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,264 @@
...
+## Prerequisites
+
+### Ironic
+
+Currently, the `baremetal` platform requires an existing Ironic environment.
+This will eventually be handled by `openshift-install`, with Ironic being
+deployed onto the bootstrap node. Until then, users of the `baremetal` platform
+should use the
+[openshift-metal3/dev-scripts](https://github.com/openshift-metal3/dev-scripts)
+repository to handle configuration of Ironic, and to launch the installer.
```
markmc: If I was an openshift/installer maintainer, I could probably handle the requirement to use dev-scripts to set up Ironic, but I'd be totally against a requirement to use dev-scripts to launch openshift-install itself

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,264 @@
...
+    api_vip: 192.168.111.5
+    hosts:
+      - name: openshift-master-0
+        role: master
+        bmc:
+          address: ipmi://192.168.111.1:6230
+          username: admin
+          password: password
+        bootMACAddress: 00:11:07:4e:f6:68
+        hardwareProfile: default
```
markmc, line 148: I struggle with this one, and I think you've answered it before - what are valid values for this field? How can a user find those out?

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,264 @@
...
+* The `image` section will get completely removed.
+
+```yaml
+apiVersion: v1beta4
+baseDomain: test.metalkube.org
+metadata:
+  name: ostest
+compute:
+- name: worker
+  replicas: 1
```
markmc, line 131: You've only listed 3 hosts below ... do you need a 4th for this worker?

docs/user/metal/install_ipi.md
```diff
@@ -0,0 +1,264 @@
...
+      checksum: 2b3b1e19e18627d89da400b63430d5bb
+      deployKernel: http://172.22.0.1/images/ironic-python-agent.kernel
+      deployRamdisk: http://172.22.0.1/images/ironic-python-agent.initramfs
+pullSecret: ...
+sshKey: ...
+```
+
+## Post-Install Customizations
+
+### NTP
```
markmc: It's great to have this awfulness documented, but a bit odd to have it just for baremetal IPI? Surely applicable in many UPI scenarios too?

markmc (changes requested):

markmc:
Awesome start, by the way!

docs/user/metal/install_ipi.md
stbenjam: These are the valid values: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/hardware/profile.go#L48

Not sure if there's user-facing docs for this, @dhellmann would know.  Maybe we should include it here if it's not anywhere else.

docs/user/metal/install_ipi.md
stbenjam: Workers not integrated into installer yet, it's done later by dev-scripts

docs/user/metal/install_ipi.md
dhellmann: The user shouldn't really have to set the hardware profile. Have you encountered cases where setting it is required?

stbenjam: @dhellmann How do derive the hardware profile? We're reusing BMO in the installer for this code, and we use the hardware profile to get the various settings needed (root device hints, etc). At the moment, the user has to specify the hardware profile in the install-config

dhellmann: The operator has some basic logic to try to derive the value. If that's not sufficient for the hardware platforms we have, we need to open some bugs describing where the guessing isn't right. 

In the mean time, it sounds like we need the user-facing field until we improve the guessing. There isn't any documentation for what the profiles are, yet, so adding some basic details here makes sense.

markmc: Unless we know of a case where we have to set this now, and we don't think that case can be quickly fixed, I wouldn't add this field - much easier to add it back later than to remove it

stbenjam: You can omit it, I suppose. However, the operator's controller code isn't being executed here, this is just being used to get the values to pass to terraform. I assume for real bare metal, you need to specify one of the dell profiles so root device hints are correct, don't you? Or is everyone basically using the default?

docs/user/metal/README.md
russellb: fixed

docs/user/metal/README.md
russellb: fixed

docs/user/metal/README.md
russellb: Good point.  Maybe I'll try "but requires the use of compatible hardware, as described in install_ipi.md"

docs/user/metal/install_ipi.md
russellb: done!

docs/user/metal/install_ipi.md
russellb: Good point.  I've removed "and to launch the installer".

docs/user/metal/install_ipi.md
russellb: I think most people are getting by fine just using the default.

I suspect there may be cases where we need to provide a root device hint, and this is the only way we have to do so.  It has to be configured in install-config, as this is stuff that has to get fed into Ironic from terraform.  There's no way to dynamically determine this at install time.

I'm going to document it as something that's likely to change while this is still considered experimental, though.

docs/user/metal/install_ipi.md
russellb: @markmc good point, I'll add a 4th host to represent a worker

@stbenjam I've glossed over that fact for the moment.  Maybe we should have a section at the end though that covers various pending PRs, and the workarounds in the meantime

docs/user/metal/install_ipi.md
russellb: Good point.  I'll move it to a more general location.

docs/user/metal/install_ipi.md
stbenjam: Note: A 4th host in the install-config won't be created in Ironic, it'll just be ignored.  That's something that needs to get fixed.  But at the moment it's the wrong Ironic anyway, as the data isn't transferred from bootstrap Ironic to the running cluster's Ironic.

russellb: I've added a new section that describes key bits of integration work that are still WIP

docs/user/metal/install_ipi.md
russellb: I've added a note about the `hardwareProfile` setting.  Please comment further if you feel it's not sufficient.  Thanks!

docs/user/metal/install_ipi.md
russellb: @stbenjam and yeah, I know it's not handled yet, but I think it's correct in terms of documenting what the `install-config.yaml` should look like.  Once the installer creates the `BareMetalHosts`, worker details need to be in here.

russellb:
I guess I should pull out the NTP docs and send them as a PR directly to the installer

russellb:
I moved the NTP docs to this PR: https://github.com/openshift/installer/pull/1951

dhellmann (approved):
The hardware profile stuff looks fine. We can continue to iterate if we need to add more details.

markmc (approved):

stbenjam:
Awesome thanks! I pulled this into https://github.com/openshift/installer/pull/1873.

PR #143, opened 2019-07-09
Small installer clean-ups identified during review

stbenjam:


pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -137,16 +137,17 @@ func (a *Bootstrap) Generate(dependencies asset.Parents) error {
 	platform := installConfig.Config.Platform.Name()
 	platformFilePath := fmt.Sprintf("bootstrap/%s/files", platform)
 	directory, err := data.Assets.Open(platformFilePath)
-	if directory != nil {
+	if err == nil && directory != nil {
```
markmc: I believe that directory == nil if err != nil is an invariant, so simply checking err is more idiomatic

```
if err == nil {
    ...
```

but perhaps we should be careful not to swallow unexpected errors. Untested, but I'd expect something like:

```
  if err == nil {
    ...
  else if !os.IsNotExist(err) {
      return err
  }
```

or

```
  switch {
  case err != nil && !os.IsNotExist(err):
      return err
  case err == nil:
      ...
   }
```

Meh

markmc (changes requested):

pkg/asset/ignition/bootstrap/bootstrap.go
stbenjam: Fair enough, I think `err == nil` is reasonable. How sure are you about os.IsNotExist being the only one we want to swallow? Asset is actually an `http.FileSystem`.  It does seem to return os.IsNotExist, but I'm not sure I want to blow everything up if some platform-specific directory doesn't exist when most platforms don't even use it.

markmc: Haven't looked carefully, but isn't this a vfsgen FS at runtime? That does return os.ErrNotExist it looks like ... wait, it looks like it's the only error it can return ... in which case just handling err == nil is fine I guess

stbenjam: Ok pushed an update for that

stbenjam:
Also, CI can't really run on this due to the platform config change on apiVIP that's needed in dev-scripts.  Let me break out the api vip change into a separate PR so I can prove CI passes on the rest.

https://github.com/openshift-metal3/kni-installer/pull/145

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/868/

cybertron (approved):

cybertron:
If the api vip isn't specified in install-config, it should be looked up from DNS. Since that works in our virt environments it seems like it should work in CI. Unless the fact that dev-scripts is setting the old name is a problem?

markmc:
> Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/868/

cluster came up fine, timed out waiting for worker - highly unlikely to be related to these trivial changes

Issue #144, opened 2019-07-09
Bootstrap coredns/keepalived files need additional documentation


stbenjam:
A number of files in  data/data/bootstrap/baremetal contain the configuration for coredns and keepalived.  The installer team has asked us to document the purpose of each file, could we add comments to each one explaining it's purpose?

stbenjam:
See: https://github.com/openshift/installer/pull/1873#issuecomment-509697686

russellb:
I've started a branch to try to document the bootstrap VM assets here: https://github.com/russellb/kni-installer/commits/baremetal-net-docs

russellb:
I just saw #146 as well.  It looks like what I've started is complementary, though.  @cybertron

stbenjam:
Have we added all the docs we want for this now?

russellb:
Yes, I think we can call this complete

PR #146, opened 2019-07-09
Add descriptions to baremetal bootstrap files

cybertron:
One thing that was requested in the upstream review of the baremetal
platform was a description of each of the files we're creating on
the bootstrap node. This change adds descriptions to each of those
files.


russellb:
lgtm - I've started some additional documentation on top of this, as well.  We'll track ongoing documentation in #144

PR #147, opened 2019-07-10
Start a document explaining baremetal bootstrap assets.

russellb:
This patch starts a document that tries to explain the purpose of all of the `baremetal` platform assets on the bootstrap VM.  The DNS part still needs more work, though.

Part of issue #144


hardys:
Looks like a good start - do you want to merge this as-is, then have @celebdor and/or @cybertron follow up with more DNS details in another PR?

markmc:
This is a great start, and it's probably easier to collaborate with follow on PRs

PR #148, opened 2019-07-10
Add design doc for bare metal IPI networking

markmc:
This document can cover the high-level requirements and design, while
README files closer to the relevant assets can cover the details.


markmc:
I figure we'll need a single high-level design doc that can connect together the bootstrap VM bits, other installer bits, MCO, etc.

russellb:
nice work!

PR #149, opened 2019-07-10
data/data/bootstrap/baremetal: fix broken link in readme

stbenjam:
The newline between ] and ( in markdown breaks the link, at least on
GitHub, and the correct path is docs, not doc.


stbenjam:
The path is also wrong, will push an update in a sec.

stbenjam:
Fixed

markmc:
Thanks, I forgot to check how it actually rendered!

PR #150, opened 2019-07-10
baremetal: remove platform survey and fix validations

stbenjam:
This is the outcome of a discussion on the openshift/installer PR. See https://github.com/openshift/installer/pull/1873#discussion_r301662942 for more context.

It is not expected that users will setup the baremetal platform in an
interactive way, given the number of data items we need from the user.
They will need to supply this data as an install-config.yaml instead.

This commit removes the survey code. The platform is also already
marked hidden, so the survey isn't accessible to an end-user anyway.


pkg/asset/installconfig/baremetal/baremetal.go
```diff
@@ -2,164 +2,10 @@
...
-// interface. net.Interfaces in Go does not let us know if it's a bridge, but
-// we can at least make sure an interface by that name exists.
-func interfaceValidator(ans interface{}) error {
-	return validate.Interface(ans.(string))
-}
-
-// macValidator validates if the answer provided is a valid mac address
-func macValidator(ans interface{}) error {
-	return validate.MAC(ans.(string))
+	return &baremetal.Platform{}, nil
```
stbenjam: This code is unreachable since the platform is hidden anyway, we could really delete it entirely.  I'm not sure what the best approach is. If we'll ever support the TUI, maybe it makes sense to leave it stubbed.

I could see if we ever use discovery the TUI becomes more usable, as a user would just be able to select which bare metal hosts to use for the control plane, for example.

markmc:
It's not a huge worry, but it's worth seeing what happens with an install-config that only specifies that baremetal platform and nothing else - where does it blow up? Can we catch it earlier and give a more friendly exit?

i.e. detect missing required stuff from install-config and error out instead of asking survey questions

pkg/asset/installconfig/baremetal/baremetal.go
markmc: If we don't validate install-config data ... that's fascinating

In general, +1 to just deleting unused code even if we might want to bring it back in future

stbenjam:
> It's not a huge worry, but it's worth seeing what happens with an install-config that only specifies that baremetal platform and nothing else - where does it blow up? Can we catch it earlier and give a more friendly exit?
> 
> i.e. detect missing required stuff from install-config and error out instead of asking survey questions

We get that detection already, for example if you omit required fields (e.g. even all of them, with a platform `baremetal: {}`), it reports the first one (which I realize isn't ideal but doesn't seem limited to our platform):

```
level=fatal msg="failed to fetch Master Machines: failed to load asset \"Install Config\": invalid \"install-config.yaml\" file: platform.baremetal.nodes: Invalid value: []*baremetal.Host(nil): bare metal hosts are missing"
```


Incidentally, we were using the wrong JSON keys, I fixed those validations here as well to report the right value.

markmc (approved):
Cool, makes sense

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/870/

stbenjam:
Failure was timeout, kicked off again.  http://10.8.144.11:8080/job/dev-tools/872/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/872/

PR #154, opened 2019-07-10
baremetal: Add more documentation about networking.

russellb:


data/data/bootstrap/baremetal/README.md
```diff
@@ -22,19 +22,33 @@ Relevant files:
 * **systemd/units/keepalived.service** - systemd unit file for `keepalived`.
   This runs `keepalived.sh` to generate the proper configuration from the
   template and then runs podman to launch `keepalived`.
+* **files/usr/local/bin/fletcher8** - 
+* **files/usr/local/bin/get_vip_subnet_cidr** -
```
stbenjam: Should these get descriptions? There are some already in the file, so wasn't sure:

fletcher8:

```
# Script that uses the fletcher8 algorithm to generate a hash from an input
# string. This is used to generate VRRP ids for use with Keepalived.
```

get_vip_subnet_cidr:

```
# Script to determine the network CIDR for a given VIP.
```

russellb: Sure, I'll copy those descriptions

data/data/bootstrap/baremetal/README.md
```diff
@@ -22,19 +22,33 @@ Relevant files:
...
 Relevant files:
-* files/etc/coredns/Corefile
-* files/etc/keepalived/keepalived.conf.tmpl
-* files/etc/dhcp/dhclient.conf
-* files/usr/local/bin/fletcher8
-* files/usr/local/bin/get_vip_subnet_cidr
-* files/usr/local/bin/coredns.sh
-* systemd/units/coredns.service
+* **files/etc/keepalived/keepalived.conf.tmpl** - `keepalived` configuration
+  template, includes configuration for the DNS VIP
```
hardys: Should we add a note here about the filename extension, e.g it's not .template because the templating is done via envsubst in https://github.com/openshift-metal3/kni-installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/keepalived.sh and not via the typical go templating inside the installer?

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -70,7 +70,101 @@ DNS requirements internal to the cluster as possible, leaving only a
...
+
+This configuration is important.  It ensures that whenever `etcd` resolves this
+DNS SRV record to form the cluster, all of the expected control plane hosts are
+present in the results.
+
+### DNS Resolution Post-Install
+
+After the bootstrap process completes, the DNS VIP moves to one of the control
+plane nodes.  All control plane nodes will continue to use this DNS VIP as
+their primary DNS server.
```
hardys, line 188: Should we mention how that's done, e.g via a NM prepender script, or is that documented in the MCO integration?

data/data/bootstrap/baremetal/README.md
```diff
@@ -22,19 +22,37 @@ Relevant files:
 * **systemd/units/keepalived.service** - systemd unit file for `keepalived`.
   This runs `keepalived.sh` to generate the proper configuration from the
   template and then runs podman to launch `keepalived`.
+* **files/usr/local/bin/fletcher8** - Script that uses the
+  [fletcher8](https://en.wikipedia.org/wiki/Fletcher%27s_checksum) algorithm to
+  generate a hash from an input string. This is used to generate VRRP ids for
```
markmc: based on the cluster name, in order to avoid collisions with other clusters on the same network?

data/data/bootstrap/baremetal/README.md
```diff
@@ -22,19 +22,37 @@ Relevant files:
...
+  use with Keepalived.  This is used by `keepalived.sh`
+* **files/usr/local/bin/get_vip_subnet_cidr** - Script to determine the network
+  CIDR for a given VIP.  This is used by `keepalived.sh`
```
markmc: which is in turn used to determine which local interface is on the VIP's network

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -70,7 +70,101 @@ DNS requirements internal to the cluster as possible, leaving only a
...
+2. Make use of mDNS (Multicast DNS) to dynamically discover the addresses of
+   hosts that we must resolve records for.
+
+### mdns-publisher
+
+https://github.com/openshift/mdns-publisher
+
+The `mdns-publisher` is the component that runs on each host to make itself
+discoverable by other hosts in the cluster.  It will be run by the
+`machine-config-operator`, once the following PR is complete:
```
markmc: it is not run on the bootstrap node by the installer because ...

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -70,7 +70,101 @@ DNS requirements internal to the cluster as possible, leaving only a
...
+
+https://github.com/openshift/machine-config-operator/pull/795
+
+### coredns-mdns
+
+https://github.com/openshift/coredns-mdns/
+
+The `mdns` plugin for `coredns` was developed to perform DNS lookups
+based on discoverable information from mDNS.  This plugin will resolve both the
+`etcd-NNN` records, as well as the `_etcd-server-ssl._tcp.` SRV record.
```
markmc, line 136: This begs the question of "what about api-int?" given the "internally resolvable" list above

Maybe move the api-int bit up to come before the mdns bit, especially since it's so simple?

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -70,7 +70,101 @@ DNS requirements internal to the cluster as possible, leaving only a
...
+   needed internal to the cluster.
+2. Make use of mDNS (Multicast DNS) to dynamically discover the addresses of
+   hosts that we must resolve records for.
+
+### mdns-publisher
+
+https://github.com/openshift/mdns-publisher
+
+The `mdns-publisher` is the component that runs on each host to make itself
+discoverable by other hosts in the cluster.  It will be run by the
```
markmc: On control plane machines, it advertises using etcd-NN and master-NN names, but on worker nodes only the worker-NN name is published

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -70,7 +70,101 @@ DNS requirements internal to the cluster as possible, leaving only a
...
+
+The `mdns` plugin for `coredns` was developed to perform DNS lookups
+based on discoverable information from mDNS.  This plugin will resolve both the
+`etcd-NNN` records, as well as the `_etcd-server-ssl._tcp.` SRV record.
+
+The list of `etcd` hosts included in the SRV record is based on the list of
+control plane nodes currently running.
+
+The IP addresses that the `etcd-NNN` host records resolve to comes from the
+mDNS advertisement sent out by the `mdns-publisher` on that control plane noded
```
markmc: Bit of a typo - "that control plane noded"

docs/design/baremetal/networking-infrastructure.md
markmc: I mention the master-NN and worker-NN names above in mdns-publisher

So coredns-mdns publishes those too

Why? Should we include those in the list of "internally resolvable DNS records" above?

docs/design/baremetal/networking-infrastructure.md
markmc: ++ need to cover dns-vip-prepender

markmc:
I think need to cover kubeconfig-extractor (and "clusterrc") somewhere

markmc:
This is awesome btw!

data/data/bootstrap/baremetal/README.md
russellb: added a note

docs/design/baremetal/networking-infrastructure.md
russellb: Good point.  There's no documentation of the MCO integration, though.  I envision this document being the high level docs that discuss how all the pieces together.  Detailed documentation of the specific MCO pieces should perhaps be in the MCO repository.  That's where I'd expect to document the exact piece that performs this task.

Once the MCO docs exist, we should be linking to them from this document.

What do you think?

data/data/bootstrap/baremetal/README.md
russellb: nice one, I've incorporated this

data/data/bootstrap/baremetal/README.md
russellb: added, thanks!

docs/design/baremetal/networking-infrastructure.md
russellb: Added:

```
`mdns-publisher` is not run on the bootstrap node, as there is no need for any
other host to discover the IP address that the bootstrap VM gets from DHCP.                                                                                                                                                                             
```

docs/design/baremetal/networking-infrastructure.md
russellb: I've moved `api-int` up as suggested, rewording it to not mention the "DNS VIP" since the DNS VIP will not have been discussed yet.

I intentionally left out mention of the `master-NN` and `worker-NN` names.  I know they were added because AWS had it, but I actually have no idea what they are used for, or if they're necessary at all.  Looking at the AWS implementation again, I don't think they are even there anymore.

@celebdor @cybertron It looks like we only need etcd records.  We may not need `mdns-publisher` on the workers at all.  Do you know of any reason that is still needed?

docs/design/baremetal/networking-infrastructure.md
russellb: Thanks, I've added this, including a note that the master-NN and worker-NN names may no longer be necessary AFAICT ...

docs/design/baremetal/networking-infrastructure.md
russellb: fixed

russellb:
> I think need to cover kubeconfig-extractor (and "clusterrc") somewhere

maybe in a more detailed doc in the MCO repo that discusses those assets?

docs/design/baremetal/networking-infrastructure.md
```diff
@@ -32,19 +38,36 @@ The API VIP first resides on the bootstrap VM. The `keepalived`
...
+The VIP will move to one of the control plane nodes, but only after the
+bootstrap process has completed and the bootstrap VM is stopped. This happens
+because the `keepalived` instances on control plane machines are configured (in
+`keepalived.conf`) with a lower
 [VRRP](https://en.wikipedia.org/wiki/Virtual_Router_Redundancy_Protocol)
-priority. These `keepalived` instances are run as [static
-pods](https://kubernetes.io/docs/tasks/administer-cluster/static-pod/)
-and the relevant assets are [rendered by the Machine Config
+priority. This ensures that the API on the control plane nodes is fully
+functional before the API VIP moves.
```
russellb, line 49: @markmc you raised the discrepency about VRRP priorities out of band.  I have corrected the documentation to reflect what's happening here.  FYI

markmc:
Awesome progress, I think we're saying 2 follows-up needed:

1. need detailed MCO docs - covering dns-vip-prepender and kubeconfig-extractor (and "clusterrc") 
2. look into whether need the master-NNN and worker-NNN records

docs/design/baremetal/networking-infrastructure.md
cybertron: I don't  know anything specific. The addresses we're serving are the ones that were in the original requirements.

I guess we can always try removing mdns-publisher from the workers and see what breaks.

PR #155, opened 2019-07-11
Convert systemd podman services to static pods

cybertron:
We want the bootstrap node to be using static pods like the masters
do. This converts the existing systemd services to static pods.


russellb:
Does this solve any particular problems?  Can you describe them in the commit message?

If it's just about consistency, can we put this off until post-4.2?

russellb:
I see it was discussed in the baremetal assets installer review: https://github.com/openshift/installer/pull/1960#commitcomment-34252968

Since that PR was approved, it probably makes sense to hold this and submit directly to openshift/installer once the above PR merges

hardys:
Yeah it seems like this would be a good follow-up for openshift/installer, and it'd be good to avoid landing anything now that will make future rebases of this repo harder.

hardys:
I just hit some issues with the current systemd implementation ref https://github.com/openshift-metal3/kni-installer/pull/100/commits/7c7d13da490596e6b79dcb271454683af29ed7d8

I suspect those will be solved by moving to staticpods, so we should probably decide if we're going to push this upstream, or shall I propose my bugfix there?  @celebdor @cybertron

hardys:
@cybertron do you know how templating works for staticpod resources?

I ask because there are several hard-coded references in my startironic.sh.template file in https://github.com/openshift-metal3/kni-installer/pull/100 - that currently uses a similar systemd approach, and we can probably template in the IPs etc from install-config additions using the go-templating implemented in the bootstrap ignition asset code.

How can we wire install-config data into staticpod yaml definitions if/when we convert all the systemd podman things to run that way?

cybertron:
Right now we template things manually in an init container. @bcrochet is working on doing the templating via MCO directly, so he's probably the one to talk to about that.

Note that we're thinking it may be necessary to do this migration now because we need to be able to retrieve the release images for coredns and keepalived. The current plan is to move these manifests into MCO, where we will deploy them in a similar way to the manifests/bootstrap-pod-v2.yaml file that's already there.

russellb:
It seems you're taking a different direction now by integrating these bootstrap static pods into MCO.  I'm going to close this PR.  All further work (issues and PRs) should now be done against the openshift/installer repo

PR #156, opened 2019-07-12
vendor: bump terraform-provider-ironic to v0.1.5

hardys:
This includes the recent changes to enable a timeout/retry
that are required for #100

### DIFF
--- a/pkg/terraform/exec/plugins/Gopkg.lock
+++ b/pkg/terraform/exec/plugins/Gopkg.lock
@@ -788,12 +788,12 @@
   version = "v1.0.0"
 
 [[projects]]
-  digest = "1:a0886267c60262a245d6eeb0be12bf8e165026ad088bce106b9c55c5d613144f"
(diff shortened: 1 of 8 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/885/

hardys:
@stbenjam should I propose this to openshift/installer now, or should we include it as part of the series when #100 is landed/working and ready to be proposed upstream?

stbenjam:
> @stbenjam should I propose this to openshift/installer now, or should we include it as part of the series when #100 is landed/working and ready to be proposed upstream?

Should be ok to propose now, we should also get the move to the metal3 org in openshift/installer as well to reduce rebase pain.

stbenjam (approved):

stbenjam:
But I'd probably avoid landing this until the rebase is in, it probably makes a conflict.

PR #157, opened 2019-07-12
Add dns vip to baremetal platform

cybertron:
This is a user-specified VIP like the API and ingress ones, and we
need it included in the baremetal platform data the same as they are.

Co-Authored-By: Yossi Boaron <yboaron@redhat.com>

### DIFF
--- a/docs/user/metal/install_ipi.md
+++ b/docs/user/metal/install_ipi.md
@@ -137,6 +137,7 @@ controlPlane:
 platform:
   baremetal:
     apiVIP: 192.168.111.5
+    dnsVIP: 192.168.111.2
     hosts:
       - name: openshift-master-0
         role: master
--- a/pkg/types/baremetal/platform.go
+++ b/pkg/types/baremetal/platform.go
@@ -63,4 +63,7 @@ type Platform struct {
 
 	// IngressVIP is the VIP to use for ingress traffic
 	IngressVIP string `json:"ingressVIP"`
+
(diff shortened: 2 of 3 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/892/

stbenjam:
What about a default value in https://github.com/openshift-metal3/kni-installer/blob/master/pkg/types/baremetal/defaults/platform.go?

cybertron:
We currently don't require the ns1 DNS record so we can't rely on the lookup like we do for the other VIPs. I guess we could hardcode the dev-scripts address in there, but I'm not sure how useful that would be.

Although I suppose we could _try_ the ns1 lookup anyway. Worst case scenario is it doesn't exist and we're back to having no value. Unless it exists and points at the wrong address, I suppose. I'm not sure how likely that is. Maybe if someone just created an ns1 record out of habit when they create the other external records for the cluster domain? In that case it would probably point at their DNS server, which would be problematic.

russellb:
Can you re-post this to openshift/installer after this merges? https://github.com/openshift/installer/pull/1873

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/904/

russellb:
Please re-submit to openshift/installer when you're ready.  I've filed this openshift/installer issue to track this: https://github.com/openshift/installer/issues/2065

PR #158, opened 2019-07-12
Rebase to ocp/release:4.2.0-0.nightly-2019-07-12-041904

hardys:
Rebase on the latest openshift/installer now that some of our PRs have merged, and also to get an updated version of RHCOS that should unblock #100


data/data/baremetal/main.tf
```diff
@@ -7,6 +7,11 @@ provider "ironic" {
   microversion = "1.52"
 }
 
+provider "ironic" {
+  url          = var.ironic_uri
+  microversion = "1.52"
+}
+
```
hardys: This is a rebase mistake

hardys:
@stbenjam and @markmc this has been tested via https://github.com/openshift-metal3/dev-scripts/pull/672 and appears to be working, if you get time to review that would be much aprpeciated :)

data/data/baremetal/main.tf
```diff
@@ -20,6 +20,7 @@ module "bootstrap" {
 module "masters" {
   source = "./masters"
 
+  ironic_uri     = var.ironic_uri
```
stbenjam: I think this shouldn't be here, we removed it in our repo but not from openshift/installer yet

data/data/baremetal/masters/variables.tf
```diff
@@ -1,8 +1,14 @@
+variable "ironic_uri" {
+  type        = string
+  description = "URI for Ironic that bootstrap should wait for to be available"
+}
+
```
stbenjam: Same here

stbenjam (commented):
Otherwise I think this looks good

data/data/baremetal/main.tf
hardys: Ah Ok, will fix and re-test

hardys: Thanks, fixed

data/data/baremetal/masters/variables.tf
hardys: Thanks, fixed

pkg/terraform/exec/plugins/Gopkg.lock
```diff
@@ -795,6 +795,14 @@
   revision = "9f5bba54dad78c10ed572eb2541f8178a5500656"
   version = "v0.1.4"
 
+[[projects]]
+  digest = "1:a0886267c60262a245d6eeb0be12bf8e165026ad088bce106b9c55c5d613144f"
+  name = "github.com/openshift-metalkube/terraform-provider-ironic"
```
stbenjam: Ooh sorry looks like this got included too, we should only use the metal3 version, I'd remove this from Gopkg.lock and the openshift-metalkube directory in vendor

stbenjam (approved):
lgtm

stbenjam:
CI passed going to merge this.

PR #159, opened 2019-07-15
Additional small clean-ups from openshift/installer review

stbenjam:
See commits for details, but this removes FIXME's and fixes some ordering issues.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -274,17 +274,17 @@ func (m *Master) Machines() ([]machineapi.Machine, error) {
 	scheme := runtime.NewScheme()
 	awsapi.AddToScheme(scheme)
 	azureapi.AddToScheme(scheme)
+	baremetalapi.AddToScheme(scheme)
 	gcpapi.AddToScheme(scheme)
 	libvirtapi.AddToScheme(scheme)
 	openstackapi.AddToScheme(scheme)
-	baremetalapi.AddToScheme(scheme)
 	decoder := serializer.NewCodecFactory(scheme).UniversalDecoder(
 		awsprovider.SchemeGroupVersion,
 		azureprovider.SchemeGroupVersion,
+		baremetalprovider.SchemeGroupVersion,
 		gcpprovider.SchemeGroupVersion,
 		libvirtprovider.SchemeGroupVersion,
 		openstackprovider.SchemeGroupVersion,
-		baremetalprovider.SchemeGroupVersion,
 	)
 
 	machines := []machineapi.Machine{}
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -179,7 +179,6 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		}
 		azure.ConfigMasters(machines, clusterID.InfraID)
 	case baremetaltypes.Name:
-		// FIXME: baremetal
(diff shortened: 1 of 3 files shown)


stbenjam:
Don't merge until latest rebase is in, to avoid creating conflicts.

pkg/asset/rhcos/image.go
```diff
@@ -67,7 +67,6 @@ func (i *Image) Generate(p asset.Parents) error {
 		//TODO(serbrech): change to right image once available.
 		osimage = "/resourceGroups/rhcos_images/providers/Microsoft.Compute/images/rhcostestimage"
 	case baremetal.Name:
-		// FIXME: baremetal
 		osimage, err = rhcos.QEMU(ctx)
```
stbenjam, line 72: This doesn't need fixing, from @russellb 

```
using the same image as libvirt is correct for the bootstrap VM
we do have: https://github.com/openshift-metal3/kni-installer/issues/37
which covers "use the right image for masters and workers"
```

pkg/asset/machines/master.go
```diff
@@ -158,7 +158,6 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		}
 		azure.ConfigMasters(machines, clusterID.InfraID)
 	case baremetaltypes.Name:
-		// FIXME: baremetal
```
stbenjam, line 182: This was introduced in a rebase,  before we created machines for baremetal.  We create them now, so this FIXME (and the one below,I think) aren't needed.

pkg/asset/rhcos/image.go
hardys: Yeah also see https://github.com/openshift-metal3/kni-installer/pull/100 where we need to consume two different images which fixes that issue, so I bypassed this abstraction - I think removing this FIXME is reasonable.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/898/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/900/

pkg/asset/machines/master.go
markmc: The thing that bothers me now with this code is that we're not using rhcosImage - i.e. nowhere in our machines resource are we describing the bootimage we used, unlike other platforms (except libvirt)

markmc: I've filed openshift-metal3/kni-installer#160 now

Even though the baremetal code path doesn't use the rhcosImage variable right now, it bothers me that the variable contains the wrong data for masters ...

pkg/asset/rhcos/image.go
markmc: This is ... ugh ...

Some of the underlying assumptions of the `rhcos.Image` abstraction:

1. The same image is used for bootstrap, masters, and workers
2. That image information is included in the generated machines manifests and also to terraform

I think we should follow that basic idea and just have an "oh hey, our bootstrap node boot image is different!" special case

markmc (approved):
Happy to remove the FIXMEs now that openshift-metal3/kni-installer#160 tracks them and gives more detail

pkg/asset/rhcos/image.go
hardys: Yeah agree we need a better abstraction here which caters for the situation where the bootstrap and master images are different, I hacked around it in #100 but would welcome feedback on possible approaches there.

Issue #160, opened 2019-07-16
No RHCOS image in baremetal Machines and MachineSets


markmc:
In the Generate() function for masters and workers under pkgs/assets/machines every other platform except libvirt generates an RHCOS image asset and includes it in the Machines or MachineSets it generates

Why isn't the baremetal platform doing this?

(NB - currently on baremetal the generated RHCOS image is the qcow image we use with libvirt for the bootstrap VM. If we were to include bootimage info in the generated machine manifests, we would need to make it have the proper URL instead)

markmc:
Some of the underlying assumptions of the rhcos.Image abstraction:

1. The same image is used for bootstrap, masters, and workers
2. That image information is included in the generated machines manifests and also to terraform

I think we should follow that basic idea and just have an "oh hey, our bootstrap node boot image is different!" special case

markmc:
Moved to openshift/installer#2037

PR #161, opened 2019-07-16
baremetal: rename os_image to bootstrap_os_image

markmc:
Rename the terraform os_image variable to bootstrap_os_image to make it clear that this boot image is only used for the bootstrap VM.

Related to openshift-metal3/kni-installer#160

### DIFF
--- a/data/data/baremetal/main.tf
+++ b/data/data/baremetal/main.tf
@@ -11,7 +11,7 @@ module "bootstrap" {
   source = "./bootstrap"
 
   cluster_id          = var.cluster_id
-  image               = var.os_image
+  image               = var.bootstrap_os_image
   ignition            = var.ignition_bootstrap
   external_bridge     = var.external_bridge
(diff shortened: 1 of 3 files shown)


markmc:
Oh - I would have proposed this to openshift/installer but only some of the code has merged

stbenjam:
> Oh - I would have proposed this to openshift/installer but only some of the code has merged

Could we defer this just a couple of days, until the openshift/installer gets merged and then do it there? It's just that this has a mix of touching merged and unmerged code

markmc:
> > Oh - I would have proposed this to openshift/installer but only some of the code has merged
> 
> Could we defer this just a couple of days, until the openshift/installer gets merged and then do it there? It's just that this has a mix of touching merged and unmerged code

ok, sounds good

hardys:
I think this can be proposed directly to openshift/install now

markmc:
Thanks

openshift/installer#2044

PR #164, opened 2019-07-16
pkg/tfvars/baremetal: check error return for bmc accessdetails

stbenjam:
Currently, if a user provides invalid BMC credentials (i.e. malformatted
or unsupported), we won't find out until much later. This adds error
checking.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/903/

PR #166, opened 2019-07-17
docs/user/metal: worker-0 should have role worker not master

stbenjam:


stbenjam:
Replaced by https://github.com/openshift/installer/pull/2027

